PR #10, opened 2021-12-09
feat(reference resolver): Add a __referenceResolver for SavedItems

kkyeboah:
## Goal
Add a reference resolver for SavedItem entity.
Immediate use of this resolver is to allow the gateway to resolve SavedItems for a user search.

- Am I doing Apollo's __referenceResolver for federation right?

## Implementation Decisions
- Added a data loader to the context for the reference resolver.

## References

Jira ticket: INFRA-212


src/resolvers/index.ts
```diff
@@ -37,6 +38,9 @@ export const resolvers = {
   SavedItem: {
     tags: savedItemTags,
     item,
+    __referenceResolver: async (savedItem, context: IContext) => {
```
Herraj, line 41: umm I'm not sure if `__referenceResolver` is the right syntax? I believe it is `__resolveReference`?

I'm following these docs: https://www.apollographql.com/docs/federation/api/apollo-subgraph/#__resolvereference

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -0,0 +1,51 @@
+import DataLoader from 'dataloader';
+import { SavedItemDataService } from '../dataService/queryServices';
+import { IContext } from '../server/context';
+import { SavedItem } from '../types';
+
+/**
+ * Reorders saved items based on the URLs that are received to
+ * keep the response in the same order of urls as requested
+ * by the Apollo Gateway (Client API)
```
kschelonka, line 9: Ah I see, this is because the SQL query response doesn't care about ordering.

src/resolvers/index.ts
kkyeboah: thank you for catching this 🙏

bassrock: @kkyeboah do we have tests that should have caught this?

kkyeboah: @bassrock added a test for the resolver

src/dataLoader/savedItemsDataLoader.ts
bassrock: Collections had a similar bug. https://pocket.slack.com/archives/CTLKW5WCF/p1638548978165400

We have to be careful when using a dataloader with reference resolvers

src/server/context.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { expect } from 'chai';
+import { ContextManager } from './context';
+import { Knex } from 'knex';
+import DataLoader from 'dataloader';
+import { SavedItemDataService } from '../dataService/queryServices';
+import { SavedItem } from '../types';
+
+jest.mock('../dataService/queryServices');
```
bassrock, line 8: this is a new pattern. Tell me more?

bassrock (approved):
👍 just a couple random comments. This looks good though.

kschelonka (approved):

PR #55, opened 2022-01-03
feat(pagination): validating pagination inputs to receive first alone, last alone or the combination of {after, first} and {before,last}.

sri-kirsh:
## Goal 
validate pagination for combination
- first
- first/after
- last
- last/after

## feedback on
- restricted to the max first/last limit to 100. this way, we can restrict large db read at once..

Jira ticket:
https://getpocket.atlassian.net/browse/INFRA-2

### DIFF
--- a/src/resolvers/utils.ts
+++ b/src/resolvers/utils.ts
@@ -2,7 +2,7 @@
  * Returns a savedItemMap from the list of tags.
  * @param tags list of tags from which savedItem keys are generated.
  */
-import { defaultPage, PaginationInput } from '../types';
+import { defaultPage, maxPageSize, PaginationInput } from '../types';
 
 export function getSavedItemMapFromTags(tags) {
   const savedItemMap = {};
@@ -40,7 +40,7 @@ export function validatePagination(pagination: PaginationInput) {
       throw new Error('invalid before cursor');
     }
 
-    if(!pagination.last) {
+    if (!pagination.last) {
       pagination.last = defaultPage;
     }
   }
@@ -51,7 +51,7 @@ export function validatePagination(pagination: PaginationInput) {
       throw new Error('invalid after cursor');
     }
 
-    if(!pagination.first) {
+    if (!pagination.first) {
       pagination.first = defaultPage;
     }
   }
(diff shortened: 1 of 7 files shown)


kschelonka (commented):
Really thorough! Small change -- can we throw an Apollo UserInputError rather than a generic Error?

src/resolvers/utils.ts
```diff
@@ -16,3 +17,60 @@ export function getSavedItemMapFromTags(tags) {
...
+  ) {
+    throw new Error('Please set either {after and first} or {before and last}');
+  }
+
+  if (pagination.before) {
+    const before = parseInt(
+      Buffer.from(pagination.before, 'base64').toString()
+    );
+    if (before < 0) {
+      throw new Error('invalid before cursor');
```
kkyeboah, line 40: nit
```suggestion
      throw new Error('Invalid before cursor');
```

src/resolvers/utils.ts
```diff
@@ -16,3 +17,60 @@ export function getSavedItemMapFromTags(tags) {
...
+
+    if (!pagination.last) {
+      pagination.last = defaultPage;
+    }
+  }
+
+  if (pagination.after) {
+    const after = parseInt(Buffer.from(pagination.after, 'base64').toString());
+    if (after < 0) {
+      throw new Error('invalid after cursor');
```
kkyeboah, line 51: nit
```suggestion
      throw new Error('Invalid after cursor');
```

src/types/index.ts
```diff
@@ -124,6 +124,7 @@ export type SavedItemUpsertInput = {
 };
 
 export const defaultPage = 30;
+export const maxPageSize = 100;
```
kkyeboah, line 127: should these be coming from a `config` file instead of a `types` file?

kkyeboah (commented):
Looks good to me. I had a few non-blocking nits

src/types/index.ts
sri-kirsh: moved to config

Herraj (approved):

PR #77, opened 2022-01-18
chore(pagination): move pagination logic to apollo-utils

sri-kirsh:
## Goal
We added the validatePagination logic to apollo-utils so it can be shared with repos like user-list-search.
In this PR, we are removing the `validatePagination` logic in list-api and using it from apollo-utils 

Ticket: 
[INFRA-244](https://getpocket.atlassian.net/browse/INFRA-244)


src/config/index.ts
```diff
@@ -81,6 +81,6 @@ export default {
   parserVersion: process.env.PARSER_VERSION || 'v3beta',
   pagination: {
     defaultPageSize: 30,
-    maxPageSize: 100
-  }
+    maxPageSize: 100,
+  },
```
jpetto, line 85: swooping in to note that updating the linter to our [latest standards](https://github.com/Pocket/collection-api/pull/618) may solve this.

jpetto (dismissed):
this looks good to me! yay for abstracting logic.

src/resolvers/user.ts
```diff
@@ -60,11 +63,13 @@ export function savedItems(
 export async function tags(
   parent: User,
   args: {
-    pagination: PaginationInput;
+    pagination: Pagination;
   },
   context: IContext
 ): Promise<TagConnection> {
-  validatePagination(args.pagination);
+  validatePagination(args.pagination,
```
sri-kirsh, line 70: todo: should assign it back to args.pagination (or) `assign pagination = {first:default} when null` inside the [pagination module ](https://github.com/Pocket/apollo-utils/blob/main/src/pagination/validatePagination.ts#L10)

src/resolvers/user.ts
```diff
@@ -39,11 +40,13 @@ export function savedItems(
   args: {
     filter: SavedItemsFilter;
     sort: SavedItemsSort;
-    pagination: PaginationInput;
+    pagination: Pagination;
   },
   context: IContext
 ): Promise<SavedItemConnection> {
-  validatePagination(args.pagination);
+  validatePagination(args.pagination,
```
sri-kirsh, line 47: todo: should assign it back to args.pagination.

kkyeboah: objects are passed by reference in javascript, this might be ok/

sri-kirsh: ya thats what i thought too, but realized we have a logic in apollo-utils `when pagination== null, return {first:defaultSize}` and this won't get assigned back to our `args.pagination` 

https://github.com/Pocket/apollo-utils/blob/main/src/pagination/validatePagination.ts#L10

i can either do this or assign `pagination = {first: defaultSize`} in apollo-utils

jpetto (approved):
:shipit:

PR #86, opened 2022-01-21
feat (savedItems): add pagination, filtering and sorting options for savedItems under tags

sri-kirsh:
## Goal
- Add pagination, sorting and filtering options for `savedItems` under `Tag` entity.

## Implementation Decisions/feedback on:
- schema change
- Add pagination, filtering and sorting for savedItems under Tag entity
- some refactored the `SavedItemDataService` to reuse the existing pagination feature.
- added test for page, sort and filtering in tags integrations test. Didn't test for every filter/sort options as we already have  extensive testing  in savedItems-filters test 
- fix assertions for createTags and updateTags test

Jira ticket: [INFRA-39](https://getpocket.atlassian.net/browse/INFRA-39)

slack discussion: https://pocket.slack.com/archives/C01VDFM30J2/p1642713584006500


src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
```diff
@@ -122,6 +133,7 @@ describe('Mutation for Tag: ', () => {
     expect(Buffer.from(data.id, 'base64').toString()).to.equal('zeta');
     expect(data._deletedAt).to.be.null;
     expect(data._version).to.be.null;
+    expect(data.savedItems.edges[0].node.id).equals('0');
   });
 
   // This test is broken and should be fixed
```
kschelonka, line 139: is this still true?

src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
```diff
@@ -197,49 +207,49 @@ describe('Mutation for Tag: ', () => {
...
+    expect(data[0].savedItems.edges[0].node.id).equals('0');
   });
 
   it('should handle other unicode, non-ascii chars', async () => {
     const variables = {
       input: [
         { savedItemId: 0, name: 'İnanç' },
         { savedItemId: 0, name: '𡞰' },
-        { savedItemId: 1, name: '(╯°□°)╯︵ ┻━┻' },
+        { savedItemId: 0, name: '(╯°□°)╯︵ ┻━┻' },
```
kschelonka, line 237: This was intentionally for a different savedItem ID and should be reverted to cover this case

src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
```diff
@@ -255,9 +265,9 @@ describe('Mutation for Tag: ', () => {
       variables,
     });
     expect(res.errors).to.undefined;
-    expect(res.data.createTags).to.deep.equal([
-      { name: 'kamille', savedItems: [{ id: '0' }] },
-    ]);
+    const data = res.data.createTags;
+    expect(data[0].name).equals('kamille');
```
kschelonka, line 269: If you're going to change this from deep equal you should also assert length == 1

src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
```diff
@@ -271,9 +281,9 @@ describe('Mutation for Tag: ', () => {
       variables,
     });
     expect(res.errors).to.undefined;
-    expect(res.data.createTags).to.deep.equal([
-      { name: 'kamille', savedItems: [{ id: '0' }] },
-    ]);
+    const data = res.data.createTags;
```
kschelonka, line 284: same as above with the deep equal removal need to check length, or revert back to deep equal

src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
kschelonka: nvm git diff made me think this was above a different line

src/test/functional/mutationServices.test/tagsMutationService-create.integration.ts
sri-kirsh: reverted this test to have `1` for `(╯°□°)╯︵ ┻━┻` 

had it changed to 0 to easily loop through asserts on savedItems as we primarily test adding unicode tags in this test. but ya its good to retain the old test behavior.

kschelonka (approved):

PR #104, opened 2022-01-25
fix(__resolveReference): add dataloader and reference resolver to get savedItems by id. 

sri-kirsh:
## Goal 

For savedItem, add resolveReference key to include `id` in addition to `url`.

user list search PR: https://github.com/Pocket/user-list-search/pull/454

Jira ticket: https://getpocket.atlassian.net/browse/INFRA-248

Docker jobs are failing due to circleCI outage: https://pocket.slack.com/archives/CMWNWJ5A9/p1643225462012500

### Feedback on:
- i am able to query only by either list of `ids` or list of `urls`. i am not able to query with this combination:
```
_representations: [
        {
          __typename: 'SavedItem',
          id: '1',
        },
        {
          __typename: 'SavedItem',
          url: 'http://def',
        },
      ],
```
is this expected?

slack thread: https://pocket.slack.com/archives/C03QVL4SU/p1643066724353300 and https://pocket.slack.com/archives/C03QVL4SU/p1643150255413100


jpetto (dismissed):
this looks solid to me!

src/dataLoader/savedItemsDataLoader.spec.ts
```diff
@@ -1,34 +1,46 @@
...
+      isFavorite: false,
+      status: 'UNREAD',
+      isArchived: false,
+      item: {
+        givenUrl: 'dont-care-too.com',
       },
-    ];
+    },
+  ];
+  it('can reorder a list of saved items based on a given list of urls', () => {
```
kschelonka, line 33:
```suggestion
  it('can reorder a list of saved items based on a given list of ids', () => {
```

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -41,10 +41,54 @@ export async function batchGetSavedItemsByUrls(
...
+ * Reorders savedItems based on the id that are received to
+ * keep the response in the same order of savedItemIds as requested
+ * by the Apollo Gateway (Client API)
+ * @param ids
+ * @param savedItems
+ */
+export function reorderSavedItemsByIds(ids: string[], savedItems: SavedItem[]) {
+  const idToSavedItemMap = savedItems.reduce((acc, savedItem) => {
+    return {
+      ...acc,
```
kschelonka, line 53: Don't need to spread this, can just rely on mutability and assign the key to the accumulator. If fetching many IDs it will be much more efficient

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -41,10 +41,54 @@ export async function batchGetSavedItemsByUrls(
...
+      [savedItem.id]: savedItem,
+    };
```
kschelonka, line 55: acc[savedItem.id] = savedItem;
return acc;

src/resolvers/item.ts
```diff
@@ -13,7 +13,7 @@ export function savedItem(
   args,
   context: IContext
 ): Promise<SavedItem> {
-  return new SavedItemDataService(context).getSavedItemByGivenUrl(
+  return new SavedItemDataService(context).getSavedItemByGivenUrls(
```
kschelonka, line 16: shouldn't this take an array of urls?

src/dataService/queryServices/savedItemsService.ts
```diff
@@ -200,24 +200,34 @@ export class SavedItemDataService {
    * Fetch a single SavedItem via its unique URL from a user's list
    * @param givenUrl the URL of the item to fetch
    */
-  public getSavedItemByGivenUrl(givenUrl: string): Promise<SavedItem> {
+  public getSavedItemByGivenUrls(givenUrl: string): Promise<SavedItem> {
```
kschelonka, line 203:
```suggestion
  public getSavedItemByGivenUrl(givenUrl: string): Promise<SavedItem> {
```

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -41,10 +41,54 @@ export async function batchGetSavedItemsByUrls(
...
+  ).batchGetSavedItemsByGivenIds(ids);
+
+  return reorderSavedItemsByIds(ids, savedItems);
+}
+
+/**
+ * Creates a dataloader for saved items to fetch by id
+ * @param context
+ */
+export function createSavedItemsDataLoaderById(context: IContext) {
```
kschelonka, line 81: won't block on this comment, but since there are two keys for loading savedItem now it could be useful to fill both caches when loading.

See [documentation](https://github.com/graphql/dataloader#loading-by-alternative-keys) here

src/dataLoader/savedItemsDataLoader.spec.ts
sri-kirsh: done

src/dataLoader/savedItemsDataLoader.ts
sri-kirsh: done

src/dataService/queryServices/savedItemsService.ts
sri-kirsh: changed all usages

src/dataLoader/savedItemsDataLoader.ts
sri-kirsh: i tried implementing this, but coz our `batchGetSavedItemsByIds` requires `context` as one of the params and with `DataLoader` accepting only id/url and not context, this gets tricky. might have to decouple the usage of context from `batchGetSavedItemsByIds`

in the below code, `prime` doesn't apply to `savedItemByUrlDataLoader` as its a function and not DataLoader object
```
const savedItemByIdDataLoader = (context: IContext, ids: string[]) => {
  let dl = new DataLoader(async (ids: string[]) => {
    let savedItems = await batchGetSavedItemsByIds(context, ids);
    for (let item of savedItems) {
      savedItemByUrlDataLoader.prime(item.id, item);
    }
    return savedItems;
  });
  return dl;
};
```

kschelonka: Ohh, if we're not sharing dataloaders across resolvers then I don't think we're actually using dataloader cache

kschelonka: Let's not make any changes, we can update as needed in another ticket.

kschelonka (approved):

PR #111, opened 2022-01-27
fix: reuse dataloaders within same request context

kschelonka:
## Goal

We should make a dataloader for each new request (each time the context is created). Subsequent calls to the dataloader on the same context object should return the same dataloader object. This fixes a bug where a new dataloader was created every time the property was accessed (resulting in no actual batch/cache).

Also adds a small optimization to fill both SavedItem loader caches when one is accessed (loading by ID/Url fills both loader caches)

## Implementation decisions
* Fill dataloader cache of alternate key when one is accessed (i.e. loading a SavedItem by ID fills the cache for SavedItem by url)

## References
[INFRA-258]

[INFRA-258]: https://getpocket.atlassian.net/browse/INFRA-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dataLoader/savedItemsDataLoader.ts
+++ b/src/dataLoader/savedItemsDataLoader.ts
@@ -30,7 +30,7 @@ export function reorderSavedItemsByUrls(
 export async function batchGetSavedItemsByUrls(
   context: IContext,
   urls: string[]
-) {
+): Promise<SavedItem[]> {
   const savedItems = await new SavedItemDataService(
     context
   ).batchGetSavedItemsByGivenUrls(urls);
@@ -62,7 +62,7 @@ export function reorderSavedItemsByIds(ids: string[], savedItems: SavedItem[]) {
 export async function batchGetSavedItemsByIds(
   context: IContext,
   ids: string[]
-) {
+): Promise<SavedItem[]> {
   const savedItems = await new SavedItemDataService(
     context
   ).batchGetSavedItemsByGivenIds(ids);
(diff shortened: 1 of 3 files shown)


src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -71,21 +71,29 @@ export async function batchGetSavedItemsByIds(
...
-export function createSavedItemsDataLoaderUrls(context: IContext) {
-  return new DataLoader((urls: string[]) =>
-    batchGetSavedItemsByUrls(context, urls)
-  );
+export function createSavedItemDataLoaders(
+  context: IContext
+): IContext['dataLoaders'] {
+  const byIdLoader = new DataLoader(async (ids: string[]) => {
+    const items = await batchGetSavedItemsByIds(context, ids);
+    items.forEach((item) => byUrlLoader.prime(item.url, item));
```
kkyeboah, line 90: nice. I get to see `prime` in action. so trippy that you can use a variable (`byUrlLoader`) before it's visually declared

kkyeboah (approved):
Thanks for catching and updating this!

src/dataLoader/savedItemsDataLoader.ts
kschelonka: I know, closures baby!

PR #161, opened 2022-02-24
fix(mutation): change to write client for all mutation calls

sri-kirsh:
## Goal
change to write client for all mutation calls

## Implementation Decisions
- merged the tagMutationService functions to tagDataService
- merged the savedItemsMutationService to savedItemsDataService
- inject savedItemsDataService into tagsDataService instead of creating them inside the constructor, so one request will have only one tagsDataService and savedItemsDataService in its workflow. 
     - note: data loader is having its own savedItemService object, but it shouldn't affect the mutation calls.
- By default the database is readClient. when a mutaiton function is called, it will send the writeClient from resolver-> service.
-  the create/update call in the service layer will return the savedItems instead of returning void. 

Note: commented out the test that checks if writeClient is being used in mutation in `savedItemsMutationService-upsert.integration`. this test passes but somehow sets the following db object to null for the test that runs after this. 

Jira ticket: [INFRA-303](https://getpocket.atlassian.net/browse/INFRA-303)


src/dataService/savedItemsService.ts
```diff
@@ -37,8 +39,11 @@ export class SavedItemDataService {
     [SavedItemStatus.DELETED]: 'DELETED',
     [SavedItemStatus.HIDDEN]: 'HIDDEN',
   };
-  private readDb: Knex;
+  private db: Knex;
+  private writeDb: Knex;
```
sri-kirsh, line 43: so we dump the read and write client into the constructor through the context when we create a savedItemsService.

src/dataService/savedItemsService.ts
```diff
@@ -422,6 +429,214 @@ export class SavedItemDataService {
...
+   * Update the 'favorite' attribute of an item, and the auditing fields
+   * in the table ('time_updated', etc.)
+   * @param itemId the item ID to update
+   * @param favorite whether the item is a favorite or not
+   */
+  public async updateSavedItemFavoriteProperty(
+    itemId: string,
+    favorite: boolean
+  ): Promise<void> {
+    this.db = this.writeDb;
```
sri-kirsh, line 454: any functions associated with the mutation call will switch the default db to `writeDb`...so any following getSavedItems call made with `this` object will only use write client

kschelonka: what about updating the query methods to optionally take a database connection instead (and default to the read connection in the class). I tend to try to avoid manipulating class properties within functions because it's harder to debug/see what's going on, in my opinion.

kschelonka: That way we can just explicitly use the write connection in the mutation resolvers

sri-kirsh: ya thats another way of implementing this and we thought about that during the discussion..then at resolver layer, we need to pass the write client instead of read with every call and we want to hide those details from the resolver layer (and let it handle only business logic)

kschelonka: Ok, I would prefer the other way but if you and Kelvin decided on it then I will support

kschelonka: I just really dislike class state

kschelonka: My alternative suggestion is that the update method should return the data required by the resolver. The only reason it's separated is because the classes weren't merged together. However that is no longer the case.

sri-kirsh: had a discussion with kat and kelvin: https://pocket.slack.com/archives/C034FB364LC/p1645815901665619

changed the resolver to pass the databaseClient object to service, and also modified the upsert/update functions to return the mutation return object.

src/test/functional/mutationServices.test/testServerUtil.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function getServer(
+  userId: string,
+  readClient: Knex,
+  writeClient: Knex,
+  eventEmitter: ItemsEventEmitter
+) {
+  return new ApolloServer({
+    schema: buildFederatedSchema({ typeDefs, resolvers }),
+    context: () => {
```
kschelonka, line 17: non-blocking, but if we're separating this out we should add the plugins to get rid of the annoying default apollo stuff

e.g. PluginInlineTraceDisabled, UsageReportingDisabled (idk exactly what they're called but similar)

src/test/functional/mutationServices.test/savedItemsMutationService-upsert.integration.ts
```diff
@@ -601,9 +575,44 @@ describe('UpsertSavedItem Mutation', () => {
         });
         expect(eventTracker.callCount).to.equal(0);
       });
+
+      //this test passes.
```
kschelonka, line 579: given we already have tests to add this item, we could just spy on the writeDb and check that it called using the server that works (non-null db).

src/resolvers/mutation.ts
```diff
@@ -99,13 +90,12 @@ export async function updateSavedItemFavorite(
...
-  await new SavedItemMutationService(context).updateSavedItemFavoriteProperty(
+  const savedItemService = createSavedItemDataServiceWithWriteDbClient(context);
+  const savedItem = await savedItemService.updateSavedItemFavoriteProperty(
     args.id,
     true
   );
-  const savedItem = await new SavedItemDataService(context).getSavedItemById(
-    args.id
-  );
+  //const savedItem = savedItemService.getSavedItemById(args.id);
```
kschelonka, line 98:
```suggestion
```

src/resolvers/mutation.ts
```diff
@@ -121,13 +111,12 @@ export async function updateSavedItemUnFavorite(
...
-  await new SavedItemMutationService(context).updateSavedItemFavoriteProperty(
+  const savedItemService = createSavedItemDataServiceWithWriteDbClient(context);
+  const savedItem = await savedItemService.updateSavedItemFavoriteProperty(
     args.id,
     false
   );
-  const savedItem = await new SavedItemDataService(context).getSavedItemById(
-    args.id
-  );
+  //const savedItem = await savedItemService.getSavedItemById(args.id);
```
kschelonka, line 119:
```suggestion
```

src/dataService/savedItemsService.ts
```diff
@@ -37,18 +39,26 @@ export class SavedItemDataService {
...
     FAVORITED_AT: 'favoritedAt',
     ARCHIVED_AT: 'archivedAt', // this is a derived field
   };
 
-  constructor(context: IContext) {
-    this.readDb = context.db.readClient;
+  constructor(
+    context: IContext,
+    db: Knex = context.db.readClient
+    //note: for mutations, please pass the writeClient,
```
kschelonka, line 56: Let's add this comment to the docstring of the class

src/test/functional/mutationServices.test/savedItemsMutationService-upsert.integration.ts
sri-kirsh: nice idea! looking into if its possible to spyOn object that gets a parameter.

tried something like this to make sure writeDb is called

```
  const db = writeClient(); //passed to apollo server
  const writeDbSelectSpy = jest.spyOn(db('list'), 'select');
  expect(writeDbSelectSpy.mock.calls.length).to.be.greaterThan(1);
```

The call `jest.toBeCalled() and jest.toHaveBeenCalled()` doesn't work with our test framework..
https://codewithhugo.com/jest-fn-spyon-stub-mock/

src/test/functional/mutationServices.test/testServerUtil.ts
sri-kirsh: just refactored the existing code that was repeatedly creating new apollo server for every test, didnt refactor the test that was calling the `getServer` call from server.ts. 

disabled `PluginInlineTraceDisabled` , usage reporting is enabled only if we provide a `GRAPH_KEY` to the server: https://www.apollographql.com/docs/apollo-server/api/plugin/usage-reporting/ 

data from inlineTracing plugin seems to be used by gateway, do you know how enabling or disabling them affect the test server?

src/resolvers/mutation.ts
sri-kirsh: done

src/resolvers/mutation.ts
sri-kirsh: done

src/test/functional/mutationServices.test/testServerUtil.ts
kschelonka: If you don't explicitly disable PluginInlineTrace, then you get a console message about missing the GRAPH_KEY when it tries to send error/usage data. Since it's just about console statements that's why it's non-blocking.

src/test/functional/mutationServices.test/savedItemsMutationService-upsert.integration.ts
kschelonka: Jest isn't great, but sinon pretty much always works :)

src/test/functional/mutationServices.test/testServerUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function getServer(
+  userId: string,
+  readClient: Knex,
+  writeClient: Knex,
+  eventEmitter: ItemsEventEmitter
+) {
+  return new ApolloServer({
+    schema: buildFederatedSchema({ typeDefs, resolvers }),
+    plugins: [ApolloServerPluginInlineTrace],
```
kschelonka, line 18:
```suggestion
    plugins: [ApolloServerPluginInlineTraceDisabled()],
```

src/test/functional/mutationServices.test/testServerUtil.ts
```diff
@@ -0,0 +1,35 @@
+import { Knex } from 'knex';
+import { ItemsEventEmitter } from '../../../businessEvents';
+import { ApolloServer } from 'apollo-server-express';
+import { ApolloServerPluginInlineTrace } from 'apollo-server-core';
```
kschelonka, line 4:
```suggestion
import { ApolloServerPluginInlineTraceDisabled } from 'apollo-server-core';
```

src/test/functional/mutationServices.test/testServerUtil.ts
kschelonka: FYI - I fixed the plugins. You have to import the `Disabled` one and call it as a function to disable it (e.g. `PluginInlineTraceDisabled()`)

kschelonka (dismissed):
won't block on the commented out test, but I recommend we fix or remove soon

src/resolvers/mutation.ts
```diff
@@ -362,7 +353,10 @@ export async function deleteTag(
   args: { id: string },
   context: IContext
 ): Promise<string> {
-  await new TagMutationService(context).deleteTagObject(args.id);
+  await createTagDataServiceWithWriteDatabaseClient(
```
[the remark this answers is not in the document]
sri-kirsh: ya moved.

src/test/functional/mutationServices.test/testServerUtil.ts
sri-kirsh: thanks! added `ApolloServerPluginUsageReportingDisabled`

src/dataService/savedItemsService.ts
```diff
@@ -422,6 +436,217 @@ export class SavedItemDataService {
...
+  private listItemUpdateBuilder(): Knex.QueryBuilder {
+    return this.db('list')
+      .update({
+        time_updated: SavedItemDataService.formatDate(new Date()),
+        api_id_updated: this.apiId,
+      })
+      .andWhere('user_id', this.userId);
+  }
+
+  public static createSavedItemDataServiceWithWriteDbClient(context: IContext) {
```
kkyeboah, line 649: please don't hate me. since it's in the same class now, could we update the name to `write` or `getWriteDbClient` or `writeDbClient`.
so `const savedItemService = SavedItemDataService.write()` for example.

100% also feel free to ignore this, I won't block on this nit

src/test/functional/mutationServices.test/savedItemsMutationService-upsert.integration.ts
kkyeboah: I agree that we should either remove this commented out test or fix.

kkyeboah (dismissed):

src/test/functional/mutationServices.test/savedItemsMutationService-upsert.integration.ts
sri-kirsh: sinon also have a similar behavior, 

```
 const spy = await sinon.spy(db('list'), 'insert');
 expect(spy.called).greaterThan(0);
 spy.resetHistory();
```

so tried spying on `insert` or `select` or `queryContext` and the assertions all failed with 0 calls.

https://github.com/knex/knex/issues/3775 might be related to this issue, we don't have `context` as they described,

i am also confused by why the test commented out would make the following test suite fail.

sri-kirsh:
https://github.com/Pocket/list-api/pull/161#discussion_r816340410

> won't block on the commented out test, but I recommend we fix or remove soon

ya agree with both that it needs to be fixed, i am worried if we remove this - we might forget the case of validating the case where the writeClient is not used, i like the idea of spying on the writeClient but it doesn't work with knex

src/dataService/savedItemsService.ts
sri-kirsh: done

kkyeboah (approved):

PR #187, opened 2022-03-05
fix(performance): slow query pagination

kschelonka:
## Goal
Fix the performance with slow queries, particularly for the cross-db joins. Uses temp tables to leverage in-memory speedups.

Also includes some bugfixes to our original pagination not handling cursor key collisions well (this should mean the `favoritedAt` sort bug is fixed).

Honestly this is a bit janky, but I feel pretty confident due to the existing and new test coverage. This code is temporary and just to unblock our clients.

I put the methods in a different class because... well... it was just so much code. and I didn't want to potentially break something else.

- Yeet the pagination library and write own methods, which properly handles not having a unique, sequential non-null value we can reliably use as a cursor
- Update some incorrect test assertions
- Fix totalCount response

current local performance (benchmarking article filter on a 50k list):
    
    first: average 66.2 ms over 20 trials (min: 58, max: 136)
    first/after: average 67.95 ms over 20 trials (min: 65, max: 85)
    last: average 60.35 ms over 20 trials (min: 57, max: 66)
    last/before: average 65.9 ms over 20 trials (min: 62, max: 68)

previous performance:

    first: average 390.55 ms over 20 trials (min: 373, max: 492)
    first/after: average 379.85 ms over 20 trials (min: 371, max: 400)
    last: average 382.75 ms over 20 trials (min: 373, max: 404)
    last/before: average 386.25 ms over 20 trials (min: 372, max: 400) 

- anything, with the caveat that this is temporary code

## Implementation Decisions
- Roll my own pagination since the knex pagination library (a) didn't handle our data well and (b) couldn't do the temp tables easily

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-341

Documentation:
* Project doc


src/test/functional/queryServices.test/savedItems.integration.ts
```diff
@@ -429,13 +429,12 @@ describe('getSavedItems', () => {
...
         sortOrder: 'DESC',
-        expectedUrls: ['http://abc', 'http://ijk'],
+        expectedUrls: ['http://ijk', 'http://def'],
       },
       {
         sortBy: 'ARCHIVED_AT',
-        // Note that this will put non-archived items first (since they are set to time 0)
         sortOrder: 'ASC',
-        expectedUrls: ['http://def', 'http://ijk'],
+        expectedUrls: ['http://abc', 'http://def'],
```
kschelonka, line 437: this was incorrect before

src/test/functional/queryServices.test/savedItems.integration.ts
```diff
@@ -429,13 +429,12 @@ describe('getSavedItems', () => {
       {
         sortBy: 'ARCHIVED_AT',
         sortOrder: 'DESC',
-        expectedUrls: ['http://abc', 'http://ijk'],
+        expectedUrls: ['http://ijk', 'http://def'],
```
kschelonka, line 432: this was incorrect before

src/types/index.ts
```diff
@@ -13,7 +13,6 @@ export type PageInfo = {
 
 export type SavedItemConnection = {
   edges: SavedItemEdge[];
-  nodes: SavedItem[];
```
bassrock, line 16: @kschelonka shouldn't this stay?

sri-kirsh: no the savedItem is inside edges.node, so we don't need it here

kschelonka: yeah this was a mistake in the schema

src/test/functional/queryServices.test/savedItems.integration.ts
kschelonka, line 432: this now correctly uses the itemId as the 'fallback' for stable sort on timestamp collisions

src/dataService/listPaginationService.ts
```diff
@@ -0,0 +1,657 @@
...
+
+  /**
+   * Wrapper for creating temp tables; call this to make sure that the temp
+   * tables created are pushed into the the instance value for `tablesCreated`,
+   * so they can be cleaned up before the transaction exits.
+   * @param trx Open transaction object
+   * @param tableName the name of the temp table created
+   */
+  private async createTempTable(trx: any, tableName: string): Promise<any> {
+    this.tablesCreated.push(tableName);
```
kkyeboah, line 184: 👀

kkyeboah (approved):
Walked through the code with Kat. Thank you for taking this on, Kat!

The test coverage missing in this PR appears to be taken care of by existing tests.

Need to fix merge conflict and a test is failing on closing a DB connection, which I find unrelated, rerun to verify.  Happy to re-approve after both are sorted, just let me know.

kkyeboah (approved):

src/dataService/listPaginationService.ts
```diff
@@ -0,0 +1,657 @@
...
+    trx: Knex.Transaction,
+    baseQuery: Knex.QueryBuilder,
+    insertStatement: string,
+    cursor: string,
+    pagination: PaginationInput,
+    sort: SavedItemsSort
+  ) {
+    const pageSize = pagination.first ?? pagination.last;
+    // Since we don't have a unique sequential column for cursor-based pagination
+    // We have to get the old cursor element + any colliding keys
```
sri-kirsh, line 299: quick question on `collission`. could you give me an example of collission with cursor

how i am understanding is: the cursor is `itemId+epoc` but for the single user. say the sort column is `updatedAt`, the cursor is `itemId+updatedAt`, and this might not collide with other item in the list as they have a different itemId.

sri-kirsh (approved):

src/dataService/listPaginationService.ts
kschelonka: sure! let's do `favoritedAt` sorted `asc`.

let's say the cursor decodes to `5_*_null`.

if there weren't any collisions (say itemId 5 was the only null value in the db), then we'd have a sort something like

```
5 | null
2 | <timestamp>
1 | <timestamp>+n // some later timestamp
6 | <timestamp>+m
...
```

This is good. The first row contains the cursor value, which is what we expect.

but if there's collisions, then our sort falls back on itemId, ascending:

```
1 | null
2 | null
5 | null
6 | null
...
```

then the first row is *not* our cursor value. And we can't just include a filter by itemId, because itemId sequence has no sequential relationship to any of the epoch timestamps. So we have to get all the collisions. Then in this case we'd strip off the first two rows, so now the cursor is the first row. Then we'd decide if we need to pull in more data or not (in the case where there are many collisions, we might have the entire page or more already).

Does that help?

PR #198, opened 2022-03-10
feat(saveditems): states filter

kschelonka:
## Goal
Add a states filter so clients can pass an array of statuses for filters, like an OR condition.

- Schema and types
- Add tests
- Update filter query

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-366


jpetto (commented):
i am mostly ignorant of how this api is used so maybe ignore me, but i have some annoying little questions 🙃

schema.graphql
```diff
@@ -254,6 +254,10 @@ input SavedItemsFilter {
   Optional, filter to get user items based on status.
   """
   status: SavedItemStatusFilter
+  """
+  Optional, filters to get user items based on multiple statuses (OR operator)
+  """
+  states: [SavedItemStatusFilter]
```
jpetto, line 260: would there be a problem if a request is made with both `status` and `states`? should `status` be deprecated in favor of `states`? why `states` and not `statuses`?

kschelonka: if someone made a request with `states` and `status`, `status` would win since it's stricter. IMO that's up to the user to use correctly. 

e.g. if it was status=UNREAD and states=[UNREAD,ARCHIVED] The sql would evaluate to:

```
... 
AND status = 0
AND status IN (0, 1)
```

> should status be deprecated in favor of states
I don't have a strong opinion on this. Technically deprecating an input field is not yet supported in the graphql spec, but it will be soon (see https://github.com/graphql/graphql-spec/pull/805). We could put the deprecation notice in the field description in the meantime.

> why states and not statuses
it was the schema change written in the ticket, so I assumed it came out of discussion with clients already (cc @kkyeboah)

kschelonka: I kind of like statuses better.

jpetto: yeah, i feel the strongest about that discrepancy. if both are operating on the same enum, i feel they should match. `status/statuses` or `state/states`.

jpetto: i feel second most strongly about deprecating `status`. if the same thing can be accomplished through the new filter, it seems cleaner to move forward with only the new filter.

BUT i have not been involved in any client discussions so i'm just the peanut gallery here.

kschelonka: Changes made

jpetto (approved):
🤩

PR #201, opened 2022-03-10
feat(filters): update contentType filters

kkyeboah:
## Goal

Update `contentType` filter per https://github.com/Pocket/spec/pull/134

## References

Jira ticket: INFRA-367

### DIFF
--- a/src/dataService/listPaginationService.ts
+++ b/src/dataService/listPaginationService.ts
@@ -546,11 +546,33 @@ export class ListPaginationService {
       'list.resolved_id',
       'readitla_b.items_extended.extended_item_id'
     );
-    if (contentType == 'VIDEO') {
-      baseQuery.where('readitla_b.items_extended.video', 1);
-    } else {
-      baseQuery.where('readitla_b.items_extended.is_article', 1);
+
+    switch (contentType) {
+      // Deprecated
+      case 'VIDEO':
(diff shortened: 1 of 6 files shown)


src/dataService/listPaginationService.ts
```diff
@@ -540,11 +540,33 @@ export class ListPaginationService {
...
+      // Deprecated
+      case 'VIDEO':
+        baseQuery.where('readitla_b.items_extended.video', 1);
+        break;
+      // Deprecated
+      case 'ARTICLE':
+        baseQuery.where('readitla_b.items_extended.is_article', 1);
+        break;
+      case 'IS_EXTERNAL':
+        baseQuery.where('readitla_b.items_extended.is_article', 0);
```
kschelonka, line 554: sorry if this was answered elsewhere: does this mean all videos/images are considered 'external'?

sri-kirsh: `Item is an un-parsable page and will be opened externally` does is_external means the clients will parse it in their end? 

we already have unmarseable right? how is this different?

kkyeboah: `IS_EXTERNAL` means that the parser did not parse it for some reason and it sets `is_article=0` for the given item. Clients open these articles in a web view instead of the native article view

PR #263, opened 2022-04-22
fix(temp tables): Ensure temp tables do not exist before creating

kkyeboah:
## Goal
Drop temp tables if they exist before creating them.

The root cause is unclear since the temp tables are dropped after use. Dropping them before creating ensures that we don't hit this issue.

## References

Sentry: https://sentry.io/organizations/pocket/issues/3084150129/?project=5761820&query=is%3Aunresolved

### DIFF
--- a/src/dataService/listPaginationService.ts
+++ b/src/dataService/listPaginationService.ts
@@ -47,10 +47,12 @@ type SavedItemResult = Omit<SavedItem, 'item' | 'tags'>;
 class Sort {
   public readonly order;
   public readonly column;
+
   constructor(sort: SavedItemsSort) {
     this.order = sort?.sortOrder.toLowerCase() ?? 'desc';
     this.column = sort?.sortBy ?? 'CREATED_AT';
   }
+
   get opposite() {
     return this.order === 'desc' ? 'asc' : 'desc';
   }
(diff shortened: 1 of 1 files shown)


src/dataService/listPaginationService.ts
```diff
@@ -193,11 +186,12 @@ export class ListPaginationService {
    */
   private dropTempTables(trx: Knex.Transaction): any {
     return Promise.all(
-      this.tablesCreated.map((tableName) =>
-        trx.raw(`DROP TEMPORARY TABLE \`${tableName}\``)
+      ListPaginationService.TEMP_TABLES.map((tableName) =>
+        trx.raw(`DROP TEMPORARY TABLE IF EXISTS \`${tableName}\``)
```
kkyeboah, line 190: Change to drop if exists

src/dataService/listPaginationService.ts
```diff
@@ -603,10 +596,10 @@ export class ListPaginationService {
     }
     const { totalCount, pageResult } =
       await this.context.db.readClient.transaction(async (trx) => {
-        await this.createTempTable(
-          this.listTempTableQuery(trx),
-          this.tempTable
-        );
+        // Drop temp tables if exists.
+        await this.dropTempTables(trx);
```
kkyeboah, line 600: drop before creating

src/dataService/listPaginationService.ts
```diff
@@ -628,7 +621,6 @@ export class ListPaginationService {
           pagination,
           sort
         );
-        await this.dropTempTables(trx);
```
kkyeboah, line 631: dropped after use

kschelonka (approved):

PR #265, opened 2022-04-25
feat(mutations): set db connection to write for all mutations

sri-kirsh:
## Goal

Use a consistent db connection (read / write) throughout the resolver chain for a given request.

For example, for the following mutation: 
```graphql
mutation CreateTags {
  createSavedItemTags({ savedItemId: 1, tags: ["this", "one"]}) {
    id
    tags {
      name  
    }
  }
}
```
`tags` is not resolved as part of the saved item, but through a separate resolver in the chain. In the original design, this would mean that the `tags` resolver will use a read db connection. The problem with using a read db connection after a mutation is that we lose strong consistency in the response to the client due to replication lag. 

The work in this PR ensures that for a mutation, every downstream resolver, uses the same connection i.e. the write db connection that was used to perform the mutation. This way we have a true representation of the state of a savedItem in the response to clients.

- Implementation

## Implementation Decisions
- Create a higher-order function that returns a resolver function for mutations:
  - Allows us to set the db connection in the context at the root of a mutation/query: ensures that every resolver in the chain uses the same connection

## References

Jira ticket: INFRA-433

### DIFF
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -25,6 +25,24 @@ import { tagsSavedItems } from './tag';
 import { SavedItem } from '../types';
 import { IContext } from '../server/context';
 
+/**
+ * wrapper function to change context database to writeDb connection
+ * @param mutate gets the mutation callback functions
+ * returns a function that changes the db to writeDb and calls the mutation callBack
+ */
+export function executeMutation<Args, ReturnType>(
+  mutate: (parent, args: Args, context: IContext) => Promise<ReturnType>
+): (parent, args: Args, context: IContext) => Promise<ReturnType> {
+  return async function (
+    parent,
+    args: Args,
+    context: IContext
+  ): Promise<ReturnType> {
+    context.db.client = context.db.writeClient;
+    return mutate(parent, args, context);
+  };
+}
+
 export const resolvers = {
   ItemResult: {
     __resolveType(savedItem: SavedItem) {
@@ -55,7 +73,7 @@ export const resolvers = {
     savedItems: tagsSavedItems,
   },
   Mutation: {
-    upsertSavedItem,
+    upsertSavedItem: executeMutation(upsertSavedItem),
     updateSavedItemFavorite,
     updateSavedItemUnFavorite,
     updateSavedItemArchive,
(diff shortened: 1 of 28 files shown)


src/server/context.ts
```diff
@@ -17,6 +17,7 @@ export interface IContext {
   apiId: string;
   userIsPremium: boolean;
   db: {
+    client?: Knex;
```
kkyeboah, line 20: this will not be optional, the default will be readClient()

kschelonka:
this seems like what we were looking for!

kschelonka (commented):
Looks like some tests need to be updated to accommodate the new `deleteSavedItemTags` response, but otherwise lgtm.

Side note -- I know mobile was on board with these changes, but should we ensure it's not a breaking change?

kkyeboah:
> Looks like some tests need to be updated to accommodate the new `deleteSavedItemTags` response, but otherwise lgtm.
> 
> Side note -- I know mobile was on board with these changes, but should we ensure it's not a breaking change?

@kschelonka I left out any of the tag updates to keep this to just the db connection changes.

src/resolvers/index.ts
```diff
@@ -70,3 +71,34 @@ export const resolvers = {
     deleteTag,
   },
 };
+
+// Wrap mutations with executeMutation to update the db connection to write
+resolvers.Mutation = Object.keys(resolvers.Mutation).reduce(
```
sri-kirsh, line 76: this would be automatically executed without a call as its in global scope right?

src/resolvers/index.ts
```diff
@@ -70,3 +71,34 @@ export const resolvers = {
...
+  (mutations: any, mutationName) => {
+    return {
+      ...mutations,
+      [mutationName]: executeMutation(resolvers.Mutation[mutationName]),
+    };
+  },
+  {}
```
sri-kirsh, line 83: nit: we can remove line 83 `{}`

src/resolvers/index.ts
kkyeboah: correct

src/resolvers/index.ts
kkyeboah: looks like you answered your question

sri-kirsh:
LGTM and good to go

src/resolvers/index.ts
kschelonka: nice, I like that this dries up the code a bit

kschelonka (approved):
🌮

PR #270, opened 2022-04-27
feat(tags): implement replaceSavedItemTags mutation

kkyeboah:
## Goal

Implement replaceSavedItemTags per https://github.com/Pocket/spec/blob/main/source/client-api/list/schema.graphql.

Related spec PR: https://github.com/Pocket/spec/pull/143

- Implementation

## Implementation Decisions
- Reused existing code for deleting and creating tags

## References

Jira ticket: INFRA-437

### DIFF
--- a/src/dataService/tagDataService.ts
+++ b/src/dataService/tagDataService.ts
@@ -240,16 +240,9 @@ export class TagDataService {
    * Note: does not check to ensure that the item being tagged
    * is actually in the user's list (no foreign key constraint).
    * @param tagInputs
-   * @param beforeInsert optional callback function to execute a query or statement before inserting tags
    */
-  public async insertTags(
-    tagInputs: TagCreateInput[],
-    beforeInsert?: (trx: Knex.Transaction) => Promise<void>
-  ): Promise<void> {
+  public async insertTags(tagInputs: TagCreateInput[]): Promise<void> {
     await this.db.transaction(async (trx: Knex.Transaction) => {
-      if (beforeInsert) {
-        await beforeInsert(trx);
-      }
       await this.insertTagAndUpdateSavedItem(tagInputs, trx);
       await this.usersMetaService.logTagMutation(new Date(), trx);
     });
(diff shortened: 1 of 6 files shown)


src/resolvers/mutation.ts
```diff
@@ -385,3 +386,37 @@ export async function updateTag(
...
+  root,
+  args: { input: SavedItemTagsInput[] },
+  context: IContext
+): Promise<SavedItem[]> {
+  const savedItems = await new TagDataService(
+    context,
+    new SavedItemDataService(context)
+  ).replaceSavedItemTags(args.input);
+
+  const savedItemTagsMap = args.input.reduce((savedItemTags, input) => {
```
sri-kirsh, line 406: having a similar implementation in createSavedItemTags and thinking about this case for emitting snowplow events.

what happens if there are duplicates, e.g
```
[
{savedItemIdA, [{tagA, tagA}]},
]
```

i see that we remove duplicates in insertTags, but we can change tags array as `set` before emitting events

src/resolvers/mutation.ts
```diff
@@ -385,3 +386,37 @@ export async function updateTag(
...
+    return {
+      ...savedItemTags,
+      [input.savedItemId]: input.tags,
+    };
+  }, {});
+
+  for (const savedItem of savedItems) {
```
sri-kirsh, line 413: also looking  at this case :
```
[
  {savedItemIdA, [{tagA, tagB}]},
  {savedItemIdA, [{tagA, tagB}]},
  {savedItemIdA, [{tagA, tagC}]},
]
```
for this case,  i guess its ok to take the last record and ignore all previous one coz its a `replace`

src/dataService/tagDataService.ts
```diff
@@ -418,6 +427,41 @@ export class TagDataService {
...
+      },
+      []
+    );
+
+    const savedItemIds = savedItemTagsInput.map((input) => input.savedItemId);
+
+    await this.insertTags(tagInputs, async (trx) => {
+      await Promise.all(
+        savedItemIds.map(async (id) => {
+          await this.deleteTagsByItemId(id).transacting(trx);
```
sri-kirsh, line 455: nit: from single responsibility principle perspective, thinking why `insertTags` need to handle the callback. would it be better to let the calling function to handle these logic i.e call `await deleteTagsByItemId` first and then do `await insertTags`

kkyeboah: It would be better but I wanted to tie into the transaction for insert tags for the delete as well. I'll try something else here, I'll ping when I update

sri-kirsh (approved):

PR #271, opened 2022-04-27
feat(createSavedItemTags): mutation to bulk create multiple tags for multiple savedItems

sri-kirsh:
## Goal
implement createSavedItemTags

- able to bulk create multiple tags for multiple savedItems
 - returns the current state of all the savedItems affected in this mutation

- used existing code for insertTags
- left the existing `createTags` mutation to not break the iOS app. 
- refactored to perform deduplicate/aggregate on inputs before passing it to the dataService. keeping this pattern consistent in both createSavedItemTags and replaceSavedItemTags

ticket: https://getpocket.atlassian.net/browse/INFRA-436

### DIFF
--- a/src/dataService/tagDataService.ts
+++ b/src/dataService/tagDataService.ts
@@ -6,6 +6,7 @@ import {
   Pagination,
   SavedItem,
   SavedItemTagAssociation,
+  SavedItemTagsInput,
   SavedItemTagUpdateInput,
   Tag,
   TagCreateInput,
@@ -238,6 +239,7 @@ export class TagDataService {
    * Insert tags into the database for items in a user's list
    * Note: does not check to ensure that the item being tagged
    * is actually in the user's list (no foreign key constraint).
+   * @param tagInputs
    */
   public async insertTags(tagInputs: TagCreateInput[]): Promise<void> {
     await this.db.transaction(async (trx: Knex.Transaction) => {
@@ -416,6 +418,43 @@ export class TagDataService {
     });
 
     return await this.savedItemService.getSavedItemById(savedItemId);
+  }
+
+  /**
+   * Replaces all tags associated with a given savedItem id
+   * @param savedItemTagsInput
+   */
+  public async replaceSavedItemTags(
+    savedItemTagsInput: SavedItemTagsInput[]
+  ): Promise<SavedItem[]> {
+    const tagInputs: TagCreateInput[] = savedItemTagsInput.reduce(
+      (inputs, tagInput) => {
+        const tagInputs = tagInput.tags.map((tag) => ({
+          savedItemId: tagInput.savedItemId,
+          name: tag,
+        }));
+
+        inputs.push(...tagInputs);
+        return inputs;
+      },
+      []
+    );
+
+    const savedItemIds = savedItemTagsInput.map((input) => input.savedItemId);
+
+    await this.db.transaction(async (trx) => {
+      await Promise.all(
+        savedItemIds.map(async (id) => {
+          await this.deleteTagsByItemId(id).transacting(trx);
+        })
+      );
+      await this.insertTagAndUpdateSavedItem(tagInputs, trx);
+      await this.usersMetaService.logTagMutation(new Date(), trx);
+    });
+
+    return await this.savedItemService.batchGetSavedItemsByGivenIds(
+      savedItemIds
+    );
   }
 
   private deleteTagsByItemId(itemId: string): Knex.QueryBuilder {
--- a/src/dataService/tagDataService.ts
+++ b/src/dataService/tagDataService.ts
@@ -421,7 +421,7 @@ export class TagDataService {
 
   /**
    * Replaces all tags associated with a given savedItem id
-   * @param savedItemTagsInput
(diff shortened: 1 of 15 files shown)


jpetto (changes requested):
lgtm, but with one place we should probably be a bit more efficient.

src/resolvers/mutation.ts
```diff
@@ -300,6 +300,50 @@ export async function createTags(
...
+
+  const savedItemTagsMap = getSavedItemTagsMap(args.input);
+
+  for (const savedItem of savedItems) {
+    context.emitItemEvent(
+      EventType.ADD_TAGS,
+      savedItem,
+      savedItemTagsMap[savedItem.id]
+    );
+  }
```
kschelonka, line 342: I think this can be simplified. `savedItemIds` isn't deduplicated, so there's not really a reason to call `getSavedItemTagsMap`. You can just use `SavedItemTagsInput` directly to emit the events.

src/resolvers/utils.ts
```diff
@@ -15,3 +17,26 @@ export function getSavedItemMapFromTags(tags) {
...
+} {
+  const savedItemTagsMap = input.reduce((savedItemTags, input) => {
+    let tags = input.tags;
+    if (savedItemTags[input.savedItemId]) {
+      tags = [...savedItemTags[input.savedItemId], ...input.tags];
+    }
+    return {
+      ...savedItemTags,
+      [input.savedItemId]: [...new Set([...tags])],
+    };
```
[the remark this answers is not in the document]
kschelonka: But see my above comment -- I don't think you need to call this function at all.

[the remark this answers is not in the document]
sri-kirsh: https://getpocket.atlassian.net/wiki/spaces/PE/pages/2677932103/Tagging+WIP this should not be high volume as the request needs to be limited to 30 items per mutation. thanks for catching this - we need to probably add the constraint to reject items greater than 30 items to prevent huge db read/write as we are not paginating the returns. 

we should add this max_record constraint to other bulk mutations too

src/resolvers/mutation.ts
sri-kirsh: the goal for calling `getSavedItemTagsMap` is to remove any duplicate savedItems and duplicate tags within savedItems, so we don't emit these duplicates in the snowplow events.

i will refactor the `getSavedItemTagsMap` function to see how we can remove duplicates in-memory

kschelonka: If we want to deduplicate and aggregate the response, we should probably do it before creating the tags and requesting back the saved items

src/resolvers/utils.ts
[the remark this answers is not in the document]
kschelonka: We could do this on the `executeMutation` method?

src/resolvers/mutation.ts
sri-kirsh: oh deduplication is done inside the `insertTags` function, this one does the same thing before sending the snowplow event

src/resolvers/utils.ts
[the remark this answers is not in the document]
sri-kirsh: ya can do this in a follow up PR

src/resolvers/mutation.ts
kschelonka: I'm wrong! no bug

jpetto (dismissed):
lgtm! (but beware i do not know this repo/system 🙃 )

src/dataService/tagDataService.ts
```diff
@@ -425,22 +424,9 @@ export class TagDataService {
    * @param savedItemTagsInput
    */
   public async replaceSavedItemTags(
-    savedItemTagsInput: SavedItemTagsInput[]
+    tagInputs: TagCreateInput[]
   ): Promise<SavedItem[]> {
-    const tagInputs: TagCreateInput[] = savedItemTagsInput.reduce(
```
sri-kirsh, line 430: moved this logic to the resolver layer

src/resolvers/mutation.ts
```diff
@@ -402,21 +441,13 @@ export async function replaceSavedItemTags(
   args: { input: SavedItemTagsInput[] },
   context: IContext
 ): Promise<SavedItem[]> {
+  const savedItemTagsMap = getSavedItemTagsMap(args.input);
+  const tagCreateInputs = convertToTagCreateInputs(savedItemTagsMap);
```
sri-kirsh, line 445: aggregate/deduplicate first before even sending it to resolver. 
note: this is handled in insertTags but safe for resolver to take care of this logic.

kkyeboah (dismissed):
lgtm

src/dataService/tagDataService.ts
```diff
@@ -425,22 +424,9 @@ export class TagDataService {
    * @param savedItemTagsInput
```
sri-kirsh, line 424:
```suggestion
   * @param tagsInputs : list of tagCreateInput
```

kschelonka (approved):
🧥

PR #273, opened 2022-04-27
feat(tags): update deleteSavedItemTags response

kschelonka:
## Goal
update `deleteSavedItemTags` to return a `SavedItem` array instead of `SavedItemTagAssociation` (per https://github.com/Pocket/spec/blob/main/source/client-api/list/schema.graphql)

- 

## Implementation Decisions
- small spec update to not allow nulls in the response at all

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-438

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -589,14 +589,13 @@ type Mutation {
   """
   deleteTag(id: ID!): ID!
   """
-  Deletes the association between a Tag and a SavedItem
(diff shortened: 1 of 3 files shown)


kschelonka:
@kkyeboah  Confirmed, they are not using the tags mutations yet

PR #274, opened 2022-04-28
feat: limit argument array size for mutations

kschelonka:
## Goal
Limit argument size for mutations

- Is this the way to go? or risk of too generic (false positive)?
- We could instead do a per-mutation resolver, or pass in some kind of decorator-style argument/function currying/callback

## Implementation Decisions
- very naive array size checker in args

### DIFF
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -26,6 +26,8 @@ import { tagsSavedItems } from './tag';
 import { SavedItem } from '../types';
 import { IContext } from '../server/context';
 import { writeClient } from '../database/client';
+import { UserInputError } from 'apollo-server-errors';
+import config from '../config';
 
 const resolvers = {
   ItemResult: {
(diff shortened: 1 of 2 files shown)


src/resolvers/index.ts
```diff
@@ -99,6 +101,13 @@ export function executeMutation<Args, ReturnType>(
     context: IContext
   ): Promise<ReturnType> {
     context.dbClient = writeClient();
+    Object.entries(args).forEach(([name, value]) => {
+      if (value instanceof Array && value.length > config.app.bulkSizeLimit) {
+        throw new UserInputError(
+          `Value of ${name} argument is too large (received=${value.length}, limit=${config.app.bulkSizeLimit}).`
```
sri-kirsh, line 107: LGTM. 

here, we are limiting the input length. in case of bulk mutation, thinking if we should limit the `tags` associated with the savedItem. i guess its not needed coz our main concern with limiting to 30 is not to return a giant savedItem list as a response

src/resolvers/index.ts
```diff
@@ -99,6 +101,13 @@ export function executeMutation<Args, ReturnType>(
     context: IContext
   ): Promise<ReturnType> {
     context.dbClient = writeClient();
+    Object.entries(args).forEach(([name, value]) => {
```
sri-kirsh, line 104: nit: do we need a UT?

kkyeboah, line 104: This ignores nested fields for now, right?
Ex. `{ input: { id: 1, tags: ["dr", "dc"] } }`, we don't check the bulk size limit for `tags`

sri-kirsh: yes right now we don't... thinking if we should 💭 

how bad can a savedItem with 10k tags, and 5 such savedItem nested in a mutation input would cost us.

we anyway have to return the all the tags we just added 1tag to a savedItem with 10k another tags.

kschelonka: yeah it ignores nested fields... let me check our schema to see how much that would affect us

PR #280, opened 2022-05-04
fix: suggested tags query speed

kschelonka:
## Goal
Remove additional join from suggested tags query to restore performance to v3. Now using the same query as the v3 endpoint for suggested tags.

In practice, clients only need the `name` or `id` fields on the suggested tags response. Those fields can be returned by the same suggested tags query as v3. The join to provide the rest of the tag metadata (e.g. saved item ids, _createdAt) resulted in poor performance. Now those fields are only fetched when requested. If the values are provided on the parent of the resolver chain (as is the case for all other tags queries except suggested tags), return that value. Otherwise, fetch the tag data using a dataloader to batch and deduplicate requests.

- Add tags dataloader
- Add field level resolvers for Tag object
- Add additional integration tests for score ordering, lazy fetching, etc.
- Create a utility function to reorder dataloader results by key/value mapping

- anything

## Implementation Decisions
- lazily load all tag attributes other than `name` and `id`; resolve from parent resolver chain if it exists, otherwise fetch using dataloader

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-450


src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -35,23 +17,10 @@ export async function batchGetSavedItemsByUrls(
...
- * @param savedItems
- */
-export function reorderSavedItemsByIds(ids: string[], savedItems: SavedItem[]) {
-  const idToSavedItemMap = savedItems.reduce((acc, savedItem) => {
-    acc[savedItem.id] = savedItem;
-    return acc;
-  }, {});
-
-  return ids.map((id) => idToSavedItemMap[id]);
+  return reorderResultByKey<SavedItem, 'url'>(
```
kschelonka, line 20: consolidating to use the generic function

src/dataService/tagDataService.ts
```diff
@@ -149,16 +144,19 @@ export class TagDataService {
            STRAIGHT_JOIN \`readitla_ril-tmp\`.suggested_tags_user_grouping_tags gt ON (g.user_id = gt.user_id AND g.grouping_id = gt.grouping_id)
            WHERE ig.resolved_id = :resolvedId
            ) AS x
-           JOIN item_tags as it
```
kschelonka, line 152: the culprit of poor performance!!

src/resolvers/tag.ts
```diff
@@ -40,11 +40,24 @@ export async function tagsSavedItems(
...
+  // If the IDs are on the parent, use them
+  if (parent.savedItems != null) {
+    return savedItemDataService.getSavedItems(
+      args.filter,
+      args.sort,
+      args.pagination,
+      parent.savedItems
+    );
+  } else {
+    // Use filter to retrieve the SavedItems if IDs are not on the parent
```
kschelonka, line 52: suggested tags won't get the list of ids, but we can easily retrieve the result we need with a filter

src/test/functional/queryServices.test/tags.integration.ts
```diff
@@ -444,69 +441,183 @@ describe('tags query tests - happy path', () => {
         }
       }
     `;
+    beforeAll(async () => {
+      getTagsByNameSpy = sinon.spy(TagDataService.prototype, 'getTagsByName');
+      await db('readitla_b.item_grouping').insert([
```
kschelonka, line 446: this was just moved to a beforeAll hook

src/test/functional/queryServices.test/tags.integration.ts
```diff
@@ -444,69 +441,183 @@ describe('tags query tests - happy path', () => {
...
+    afterEach(() => {
+      getTagsByNameSpy.resetHistory();
+    });
+
+    it('return list of SuggestedTags and paginated savedItems for SavedItem, ordered by score descending', async () => {
+      const res = await server.executeOperation({
+        query: GET_SUGGESTED_TAGS_FOR_SAVED_ITEM,
+        variables,
+      });
+      const expectedTags = [
```
kschelonka, line 511: simplifying the tags assertions to use an object

src/resolvers/index.ts
```diff
@@ -56,6 +56,25 @@ const resolvers = {
...
+    // Fetching the below values in the suggested tags query (via join to `item_tags`)
+    // dramatically reduces performance; only return them if requested. If the field is
+    // already provided on the parent context, return that rather than fetching again.
+    // Otherwise, batch load with DataLoader to avoid repeat calls to the database.
+    _createdAt: (
+      parent: Tag,
+      args,
+      context: IContext
+    ): Promise<Tag['_createdAt']> =>
+      lazyLoadTagAttribute(parent, context, '_createdAt'),
```
sri-kirsh, line 71: ah nice! these values gets resolved only when mentioned in the graphql request

src/dataLoader/savedItemsDataLoader.ts
sri-kirsh: having trouble understanding why need to reorder tags and at resolver layer, how would the clients provide reorder input? coz the schema for suggested tags only has this line: https://github.com/Pocket/list-api/blob/main/schema.graphql#L119

src/test/functional/queryServices.test/tags.integration.ts
```diff
@@ -444,69 +441,183 @@ describe('tags query tests - happy path', () => {
...
+        {
+          user_id: 1,
+          grouping_id: 10631688,
+          tag: 'zebra',
+          weighted_count: 0.9211,
+          count: 2,
+        },
+        {
+          user_id: 1,
+          grouping_id: 10631689,
```
sri-kirsh, line 469: trying to understand this table - does the `grouping_id` groups similar tags? 

and what does the weighted_count and count signify? sorry for so many context question - can do quick zoom call for PR walk through if its easier

src/test/functional/queryServices.test/tags.integration.ts
```diff
@@ -444,69 +441,183 @@ describe('tags query tests - happy path', () => {
...
+
+    it('return list of SuggestedTags and paginated savedItems for SavedItem, ordered by score descending', async () => {
+      const res = await server.executeOperation({
+        query: GET_SUGGESTED_TAGS_FOR_SAVED_ITEM,
+        variables,
+      });
+      const expectedTags = [
+        {
+          id: Buffer.from('travel').toString('base64'),
+          name: 'travel',
```
sri-kirsh, line 514: i might be missing context, in this test case -  both `zebra` and `travel` already would show up as tags as they have `item_tags` association with item_id: 1,
how does suggested tags differ here?

src/dataLoader/savedItemsDataLoader.ts
sri-kirsh: having trouble understanding why need to reorder tags and at resolver layer, how would the clients provide reorder input? coz the schema for suggested tags only has this line: https://github.com/Pocket/list-api/blob/main/schema.graphql#L119

edit: synced with kelvin, we need to preserve the order for the data loader

kkyeboah (approved):

sri-kirsh (approved):
LGTM 🚀 matches what we described in the PR description

had a walkthrough with kelvin and got most of it - except `suggested_tags` grouping and other fields in the suggested_tags table.

src/test/functional/queryServices.test/tags.integration.ts
sri-kirsh: how i understood this: say if the user has 1000 tags in `item_tags` table - then we would return the closest few tags as suggested but it would be a long list to test here

PR #288, opened 2022-05-11
feat(sqsLambda): Add sqs event consumer lambda to service infra

Herraj:
## Goal

Add event consumer workflow (SQS + Lambda Target).

The Lambda is currently a no-op (just prints the message), and there is no app logic to actually send messages to the queue at this time.

Tested in dev and the integration is working.

- 

## Implementation Decisions
- Chose not to fix the initial lambda deployment issue in this PR. Will update our circleci orbs for a more general fix at another time.

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-494


.aws/src/SqsLambda.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const { sentryDsn, gitSha } = this.getEnvVariableValues();
+
+    new PocketSQSWithLambdaTarget(this, 'sqs-event-consumer', {
+      name: `${config.prefix}-Sqs-Event-Consumer`,
+      batchSize: 10,
+      batchWindow: 60,
+      sqsQueue: {
+        maxReceiveCount: 3,
+        visibilityTimeoutSeconds: 300,
```
Herraj, line 27: nit pick: should we lower this down to `150`? Asking because that's what we have for the curation migration datasync lambda

.aws/src/SqsLambda.ts
```diff
@@ -0,0 +1,76 @@
...
+      },
+      lambda: {
+        runtime: LAMBDA_RUNTIMES.NODEJS14,
+        handler: 'index.handler',
```
Herraj, line 31: This is more of a learning/clarifying question - we do have this here and we have the file `../../lambda/index.ts` which is exporting the function `handler`. How does that function get mapped / "picked up" by the lambda? I don't see a explicit path pointing to that file/function 🤔

Herraj (commented):
Approve but I can't approve as an author 😄

.aws/src/SqsLambda.ts
kschelonka: responded in slack!

kschelonka (approved):

.aws/src/SqsLambda.ts
kschelonka: hmm, this was the default we (also) copied from elsewhere. I'll leave it for now and revisit it if we need to 👍

PR #302, opened 2022-05-20
fix(ci): conditional build working and parameter fix

kschelonka:
## Goal

🥕


kschelonka:
Tested dev deployment with force-build -- both worked and the dev resources are in aws!

PR #308, opened 2022-05-24
feat(batchDelete): add batch delete endpoint

kschelonka:
## Goal
Add basic batchDelete REST endpoint with schema validation.

- the structure 

## Implementation Decisions
- Use modular routes
- Use express-validator to enforce contract
- Include a trace ID for log analysis
- Don't keep http connection open waiting for promise to resolve -- just kick off the transaction and return a 'gotchu' response
  - errors sent to cloudwatch/sentry asynchronously

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-571

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "knex": "0.95.15",
     "locutus": "2.0.16",
     "luxon": "2.3.2",
-    "mysql": "2.18.1"
+    "mysql": "2.18.1",
+    "nanoid": "3.3.4"
   },
   "devDependencies": {
     "@apollo/federation": "0.36.1",
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "aws-xray-sdk-core": "3.3.5",
     "aws-xray-sdk-express": "3.3.5",
     "dataloader": "2.0.0",
+    "express-validator": "^6.14.1",
     "graphql": "15.8.0",
     "graphql-depth-limit": "1.1.0",
     "graphql-tools": "8.2.11",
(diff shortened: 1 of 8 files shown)


src/server/routes/batchDelete.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    customSanitizer: {
+      options: (value) => {
+        return value.map((_) => parseInt(_));
+      },
+    },
+  },
+};
+
+function validate(
```
kschelonka, line 47: this can be pulled out if we want to have shared validator middleware

src/dataService/savedItemsService.ts
```diff
@@ -36,7 +37,7 @@ export class SavedItemDataService {
   private readonly userId: string;
   private readonly apiId: string;
 
-  constructor(context: IContext) {
+  constructor(context: Pick<IContext, 'dbClient' | 'userId' | 'apiId'>) {
```
kschelonka, line 40: Simplifying the type here so that I don't have to create a whole context -- only what we actually need

package.json
```diff
@@ -55,6 +57,7 @@
     "@types/luxon": "2.3.2",
     "@types/node": "16.11.33",
     "@types/sinon": "10.0.11",
+    "@typescript-eslint/eslint-plugin": "^5.26.0",
```
bassrock, line 60: oh dang. Was this required? `eslint-config` should be pulling this in automatically :(

kschelonka: oh, I was getting an error running locally. I can remove and try again

kkyeboah (approved):
nice work!

PR #336, opened 2022-06-07
fix: keep promises in memory and send to sqs at once

kschelonka:
This avoids issues with inconsistent offset pagination, which can occur if rows are being deleted as promises are resolved asynchronously.

Uses the same pattern as https://github.com/Pocket/account-data-deleter/pull/40/

### DIFF
--- a/src/server/routes/queueDelete.ts
+++ b/src/server/routes/queueDelete.ts
@@ -7,7 +7,6 @@ import config from '../../config';
 import { sqs } from '../../aws/sqs';
 import {
   SendMessageBatchCommand,
-  SendMessageBatchCommandOutput,
   SendMessageBatchRequestEntry,
 } from '@aws-sdk/client-sqs';
 import { nanoid } from 'nanoid';
@@ -86,7 +85,7 @@ export async function enqueueSavedItemIds(
   const { userId, email, status } = data;
   const limit = config.queueDelete.queryLimit;
   let offset = 0;
-  const sqsSends: Promise<SendMessageBatchCommandOutput>[] = [];
+  const sqsCommands: SendMessageBatchCommand[] = [];
   let sqsEntries: SendMessageBatchRequestEntry[] = [];
 
   const loopCondition = true;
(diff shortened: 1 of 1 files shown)


src/server/routes/queueDelete.ts
```diff
@@ -121,12 +124,25 @@ export async function enqueueSavedItemIds(
...
+            command
+          )})`;
+          Sentry.addBreadcrumb({ message });
+          Sentry.captureException(err);
+          console.log(message);
+        });
+      })
+    );
   } catch (e) {
+    // Also provide top-level logging
```
sri-kirsh, line 145: this now doesnt need a try-catch right as each promise would handle their own error?

src/server/routes/queueDelete.ts
```diff
@@ -73,7 +72,11 @@ router.post(
 );
 
 /**
- * Enqueue item IDs for deletions in batches
+ * Enqueue item IDs for deletions in batches. This is used to purge
+ * user list and tag data when a user deletes their account. Since
+ * the data to clear could be large, we don't want to keep the api
+ * connection open while it's happening. Instead these processes
+ * will happen asynchronously in the background using queues.
```
sri-kirsh, line 79: nice comment on async, will take this for account-data-deleter :D

sri-kirsh: in account-data-deleter, changing this as `Instead the generation of the sqs messages will happen asynchronously in the background.`

src/server/routes/queueDelete.ts
```diff
@@ -109,7 +112,7 @@ export async function enqueueSavedItemIds(
       );
 
       if (sqsEntries.length === config.aws.sqs.batchSize) {
-        sqsSends.push(sqsSendBatch(sqsEntries));
+        sqsCommands.push(buildSqsCommand(sqsEntries));
```
sri-kirsh, line 115: are we continuing to use `chunk`? 

5000 records of primary key comes to 20kb and batch sqs message is upto 256kb, so we removed chunk in the account-data-deleter.

sri-kirsh (approved):

PR #338, opened 2022-06-14
fix(bug): return time_read as archivedAt

kkyeboah:
## Goal
- Limit pagination maximum to 30
- Correctly use `time_read` as `archivedAt` in the `SavedItem` entity

## References

Jira ticket: INFRA-582


nina-py (approved):
🚀

PR #367, opened 2022-07-07
feat(accountDelete): app polls batch delete queue synchronously

kschelonka:
## Goal
Use app to poll the BatchDelete queue and process messages synchronously, rather than a lambda integration. This will reduce load on the database and enable us to limit the rate.

Checked permissions with policy simulator. Feeling pretty confident with the code as it's the same as AccountDataDeleter (knock on wood). This will be QA'd during the e2e testing with prod snapshot.

- rarely do I say nothing but this is a copy job of previously reviewed code mostly :) 

## Implementation Decisions
- Mostly copying accountDataDeleter code

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/BACK-642

### DIFF
--- a/src/server/main.ts
+++ b/src/server/main.ts
@@ -14,7 +14,7 @@ import {
 } from '../businessEvents';
 import queueDeleteRouter from './routes/queueDelete';
 import { BatchDeleteHandler } from '../aws/batchDeleteHandler';
-import { EventEmitter } from 'stream';
+import { EventEmitter } from 'events';
 
 //Set XRAY to just log if the context is missing instead of a runtime error
 


src/aws/batchDeleteHandler.ts
```diff
@@ -0,0 +1,171 @@
...
+   * BatchDelete queue
+   * @param message SQS Message recieved from BatchDelete queue
+   */
+  async deleteMessage(message: Message) {
+    const deleteParams = {
+      QueueUrl: config.aws.sqs.listDeleteQueue.url,
+      ReceiptHandle: message.ReceiptHandle,
+    };
+    try {
+      await this.sqsClient.send(new DeleteMessageCommand(deleteParams));
```
bassrock, line 69: i love how aws cant just do a simple `.delete()`...

src/aws/batchDeleteHandler.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  async handleMessage(body: BatchDeleteMessage): Promise<boolean> {
+    try {
+      const traceId = body.traceId ?? nanoid();
+      const dbClient = writeClient();
+      await new SavedItemDataService({
+        dbClient,
+        userId: body.userId.toString(),
+        apiId: 'backend',
+      }).batchDeleteSavedItems(body.itemIds, traceId);
```
bassrock, line 95: whats the max number that can be here?

bassrock (approved):

src/aws/batchDeleteHandler.ts
kschelonka: it's not enforced in `batchDeleteSavedItems`, although it could be... we're relying on the queue logic which uses the value of `config.queueDelete.itemIdChunkSize` which = `100` right now.

PR #383, opened 2022-07-15
fix: increase sqs visibility config and retention time

sri-kirsh:
## Goal

Increase sqs visibility timeout so the messages don't reappear in the queue quickly. we do this coz some deletion queries run upto 1000s.

increase message retention time to 14 days (max allowed)

set task count to 1 as this app is used by beta client only for now.. 

### future todo
move the batchDeleteHandler to its own ecs task. so we can scale it down to 1 without affecting the API scaling in a future PR - https://getpocket.atlassian.net/browse/INFRA-662

### DIFF
--- a/.aws/package.json
+++ b/.aws/package.json
@@ -18,7 +18,7 @@
     "node": ">=12"
   },
   "dependencies": {
-    "@pocket-tools/terraform-modules": "4.0.2"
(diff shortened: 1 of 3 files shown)


jpetto (approved):
lgtm

PR #395, opened 2022-07-19
fix: remove item_session; moving it to account-data-deleter

sri-kirsh:
## Goal
remove item_session; moving it to account-data-deleter to delete the rows by PK `id`. this table doesn't have the composite index (user_id, item_id)

### DIFF
--- a/src/dataService/savedItemsService.ts
+++ b/src/dataService/savedItemsService.ts
@@ -286,7 +286,6 @@ export class SavedItemDataService {
       'item_ads',
       'item_time_spent',
       'item_currently_reading',
-      'item_session',
       'list_extras',
       'list_meta',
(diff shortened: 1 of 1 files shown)


src/dataService/savedItemsService.ts
```diff
@@ -297,8 +296,8 @@ export class SavedItemDataService {
         try {
           await this.db(table)
             .delete()
-            .where('item_id', id)
-            .andWhere({ user_id: this.userId });
+            .where({ user_id: this.userId })
```
sri-kirsh, line 299: this ordering doesn't matter according to documentation as long as both these fields are in the same composite index.

bassrock (approved):

PR #402, opened 2022-07-22
feat: emit account-delete-complete event 

sri-kirsh:
## Goal
list-api able to emit account-delete complete event to the event bridge

### Implementation decisions:
- aws iam permission to emit event to event bus
- `emitAccountDeletionEvent` emits the event within list-api, to which the account-deletion handler listens to.
- `AccountDeletionEventHandler` emits the event to the event bridge

TODO:
- when to call the `emitAccountDeletionEvent`? - follow-up work tracked in this ticket: https://getpocket.atlassian.net/browse/INFRA-716

Ticket: https://getpocket.atlassian.net/browse/INFRA-592


src/aws/batchDeleteHandler.ts
```diff
@@ -16,7 +16,9 @@ import { setTimeout } from 'timers/promises';
 export type BatchDeleteMessage = {
   traceId: string;
   userId: number;
-  itemIds: number[];
+  itemIds?: number[];
```
sri-kirsh, line 19: made this optional so can be be re-used by emitAccountDeleteEvent, can also create a new type if it helps to have itemIds as non-optional field

.aws/src/config/index.ts
```diff
@@ -43,6 +43,7 @@ export const config = {
     sqsPermLibItemMainQueueName: `PermLib-${environment}-ItemMain`,
     unifiedEventStreamName: 'unified_event',
     databaseTz: 'US/Central',
+    eventBusName: `PocketEventBridge-${environment}-Shared-Event-Bus`,
```
kschelonka, line 46: let's make this 'default' so we don't have to change it later?

src/aws/batchDeleteHandler.ts
kschelonka: I think this has to be a required field for BatchDeleteMessage. If you want to reuse it you could make an `Omit` type, or make a base type `UserMessageBase` or something that `BatchDeleteMessage` extends?

kschelonka (approved):

src/aws/batchDeleteHandler.ts
sri-kirsh: done

.aws/src/config/index.ts
sri-kirsh: done

kschelonka (approved):

PR #419, opened 2022-08-01
fix(suggested-tags): adding in a better suggested tags query

bassrock:
## Goal

Switch the suggested tags query to be our newer one from the Web repo that does not use tables we will no longer populate https://github.com/Pocket/Web/blob/d4920d1e8a333ce3ff6bfce0971d0bfadbc40315/src/User/LegacyFunctionSuggestedTags.php#L113


src/dataService/tagDataService.ts
```diff
@@ -123,31 +123,53 @@ export class TagDataService {
...
+      TO_BASE64(tag) AS id
+      FROM (SELECT tag,
+                  sum AS score
+          FROM (SELECT m.tag,
+                        SUM(m.count) * 100 AS sum
+                FROM (SELECT it.tag,
+                              ig.grouping_id,
+                              COUNT(1) AS count
+                      FROM (SELECT tag, item_id, user_id
+                            FROM \`readitla_ril-tmp\`.item_tags
```
kschelonka, line 138: > `\``

aha

kschelonka (approved):
Admittedly I reviewed the testing more like regression/snapshot tests, with the knowledge that this query came from analytics (and we assume they did due diligence on the logic)

PR #466, opened 2022-09-01
feat(suggestedTags): Return 3 last used tags

kschelonka:
## Goal
Add query to fetch the last 3 tags used for suggested tags. 

Excludes any existing tags from suggested tags (see skipped test).

## Implementation Decisions
- ~Probably should exclude any tags already on the saved item, but depending on how suggested tags are fetched (only on first save?) maybe this isn't super important~ DONE
- Remove lazy loading of tag attributes as it is no longer necessary (suggested tags easily returns the full tag entity now)

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-735


nina-py (dismissed):

bassrock (approved):
So much more effecient!

PR #474, opened 2022-09-12
fix: filter out deleted item from data-loader on batch get by id

sri-kirsh:
## Goal
fix: filter out deleted item from data-loader on batch get by id

### Implementation decisions/feedback on:
- data loaders filter out deleted items when we retrieve by `id` with an assumption we won't have other services consuming deleted items through data loaders. only use case i can think of is `undo-delete` which is handled within list-api.

ticket: https://getpocket.atlassian.net/browse/INFRA-751


jpetto (approved):
code-wise this looks good. my only question would be client expectations. will clients expect this kind of behavior? what are the current client impacts before this change?

sri-kirsh:
>my only question would be client expectations. will clients expect this kind of behavior? what are the current client impacts before this change?

- clients don't want the deleted items through data loaders as we shouldn't be returning deleted items in the search or from any other microservice
- yes they expect this behavior as @collectedmind reported the bug

PR #526, opened 2022-10-18
fix(suggested tags): get time_updated from list table when item_tags's time_updated is null 

sri-kirsh:
## Goal
time_updated in `item_tags` table are sometimes set to null by android clients. this breaks suggested tags response as we return top 3 tags that has been updated recently and any recently updated tag with null value automatically goes to last in the list.

As a fix: we are considering list table's time_updated when the item_tag's time_updated is null. the reasoning behind this is we update savedItem's updated_At everytime we add a tag or update a tag. so even though this is not a clear fix, this can return tags close to recently updated time. 

## Implementation Decisions/feedback on
slack discussion: https://pocket.slack.com/archives/C03QVL4SU/p1666111083106169
- code change to join with list table and consider `list`'s `time_updated` when `item_tags`'s time_updated is null.
- integration test to cover this use case

Ticket: https://getpocket.atlassian.net/browse/INFRA-773


src/test/functional/queryServices.test/suggested-tags.integration.ts
```diff
@@ -193,5 +195,83 @@ describe('tags query tests - happy path', () => {
...
+      const res = await server.executeOperation({
+        query: GET_SUGGESTED_TAGS,
+        variables,
+      });
+
+      const tags = res.data?._entities[0].savedItemById.suggestedTags;
+      expect(tags.length).to.equal(3);
+      //tagC is not in top 3 as it has the lowest time_updated
+      //null date tag will get its time_updated from list table.
+      expect(tags[0].name).to.equal('tag A');
```
kschelonka, line 272: Curious about this ordering -- shouldn't it 'null date' (date3), then 'tag A' (date2), then 'tag B' (date1), for descending order by date?

sri-kirsh: ya, we do a `getTags` query with 3 matching tags.

i understood the intend of this query as to return top 3 recent tags (not in any particular order), so didn't change this logic

https://github.com/Pocket/list-api/blob/ea79076916fdbe79a74f3b9b51f1866afbe1b3ef/src/dataService/tagDataService.ts#L133

kschelonka: Ok, seems like a requirements question to me! Mind bringing it to Slack and checking with Amber?

kschelonka (approved):

PR #534, opened 2022-10-25
fix(suggestedTags): better workaround for bad data in item_tags table

kschelonka:
## Goal
Improve most recent tags query based on real use testing and discovering missing data. Use time_added field instead of time_updated field for retrieving most recent tags.

This query was tested on my account in the prod database and improved my results. It also makes sense, since time_added is the time the association was created, and we don't really have a concept of updating the association. 

- Use the time_added field as the first sort field when retrieving recently used tags -- time_added is more reliable than time_updated
- Add a fallback for _updatedAt field in Tag entity, which is almost never populated (coalesce with time_added)
- Update tests to cover more realistic cases


src/dataService/tagDataService.ts
```diff
@@ -56,7 +56,9 @@ export class TagDataService {
         this.db.raw(`TO_BASE64(tag) as id`),
         this.db.raw('GROUP_CONCAT(item_id) as savedItems'),
         this.db.raw('UNIX_TIMESTAMP(MIN(time_added)) as _createdAt'),
-        this.db.raw('UNIX_TIMESTAMP(MAX(time_updated)) as _updatedAt'),
+        this.db.raw(
+          'UNIX_TIMESTAMP(MAX(COALESCE(time_updated, time_added))) as _updatedAt'
```
sri-kirsh, line 60: reading from PR description that we don't update the `time_updated` for item_tags table: wondering if we can just read `time_added`?

kschelonka: I'm not 100% sure that it's never updated, but probably

src/dataService/tagDataService.ts
```diff
@@ -123,8 +125,10 @@ export class TagDataService {
       .whereNotIn('tag', existingTags)
       .andWhere({ 'item_tags.user_id': parseInt(this.userId) })
       //for now, considering time_updated from list table when item_tags
-      //time_updated is null.
-      .orderByRaw('COALESCE(item_tags.time_updated, list.time_updated) DESC')
+      // time_added is null.
+      // Not using item_tags.time_updated or coalescing with item_tags.time_added
```
sri-kirsh, line 129: extreme nit but just clarifying: `Not using item_tags.time_updated . Instead, coalescing with item_tags.time_added`

sri-kirsh (approved):

src/dataService/tagDataService.ts
kschelonka: This was incredibly unclear language. I can update. I'm actually saying:

Not using or coalescing the item_tag.time_updated field because it's so unreliable and not worth it

PR #538, opened 2022-10-26
fix(suggestedTags): aggregate function order by (max timestamp)

kschelonka:
## Goal

I couldn't get the tests to fail, which is why I didn't update them...

This should solve the issue with inconsistent dates.

### DIFF
--- a/src/dataService/tagDataService.ts
+++ b/src/dataService/tagDataService.ts
@@ -132,7 +132,7 @@ export class TagDataService {
       //
       // Fall back on the time the Save was last updated. This fallback
       // time may not be when the tag was added, but it's the best proxy we have.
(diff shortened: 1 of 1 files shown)


sri-kirsh:
ya i tried adding a duplicate entry on the test like this:
```
{
          user_id: 1,
          item_id: 10,
          tag: 'tag B',
          status: 1,
          time_added: date4, //highest timestamp val 
          time_updated: date4,
          api_id: 'apiid',
          api_id_updated: 'updated_api_id',
        }
```
so tag B is associated with both items 12 and 10, but the current code handles this too

sri-kirsh (approved):

kschelonka:
Yeah, I tried a lot of different things -- duplicates, old dates, collisions, etc. But I could not get it to fail. Short of dumping my user profile data (which I don't really want to do) I'm not sure what else to do!

PR #541, opened 2022-11-02
feat(batch delete): delete all itemIds passed for batchDeleteHandler 

sri-kirsh:
## Goal
delete all itemIds passed for batchDeleteHandler 

## Implementation Decisions
-  add option to delete all itemIds passed to the deleteHandler
- `itemsChunkSize` is right now 20. so we should be deleting 20 items at once.

ticket: https://getpocket.atlassian.net/browse/INFRA-779


src/test/functional/queryServices.test/suggested-tags.integration.ts
```diff
@@ -238,6 +239,16 @@ describe('tags query tests - happy path', () => {
...
+        {
+          user_id: 1,
+          item_id: 10,
+          tag: 'tag B',
+          status: 1,
+          time_added: date4,
+          time_updated: date4,
+          api_id: 'apiid',
+          api_id_updated: 'updated_api_id',
+        },
```
kschelonka, line 251: this is why the test is failing -- added more data. Probably a leftover from when you were trying to get the suggested tags test to fail?

src/test/functional/queryServices.test/suggested-tags.integration.ts
```diff
@@ -13,6 +13,7 @@ describe('tags query tests - happy path', () => {
   const date1 = new Date('2021-10-03T10:20:30.000Z');
   const date2 = new Date('2022-10-03T10:20:30.000Z');
   const date3 = new Date('2023-10-03T10:20:30.000Z');
+  const date4 = new Date('2023-10-04T10:20:30.000Z');
```
kschelonka, line 16: same here

src/test/functional/queryServices.test/suggested-tags.integration.ts
sri-kirsh: oh yes, thanks for catching that Kat - shifting branches moves the unstaged changes too 🤦

kschelonka (approved):

PR #554, opened 2022-11-15
chore: batch delete 50 records every 10 seconds

sri-kirsh:
## Goal
if the cpu load is still under 50%, we can look to batch delete 50 records every 10 secs

Ticket: https://getpocket.atlassian.net/browse/INFRA-779

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -108,10 +108,10 @@ export default {
   },
   queueDelete: {
     queryLimit: 500,
-    itemIdChunkSize: 20,
+    itemIdChunkSize: 50,
   },
   batchDelete: {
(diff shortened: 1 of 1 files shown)


src/config/index.ts
```diff
@@ -108,10 +108,10 @@ export default {
   },
   queueDelete: {
     queryLimit: 500,
-    itemIdChunkSize: 20,
+    itemIdChunkSize: 50,
   },
   batchDelete: {
-    deleteDelayInMilliSec: 20000,
```
sri-kirsh, line 114: changing this from 20 to 10 to keep it consistent with account-deletion service..

bassrock (approved):

Herraj (approved):

PR #567, opened 2022-11-17
chore: add more breadcrumbs in data loader to identify items failing

sri-kirsh:
## Goal

- we are trying to access `url` for an item that doesn't exist
- `reorderResultByKey` could set the item to undefined if the Id is not present in the database or when it got deleted during the batchRead

### Implementation details 
- add to dataloader only when the item is present
- filter the undefined values in both `batchGetSavedItemsByIds` and `batchGetSavedItemsByUrls`
- added test to make sure undefined values and deleted items are filtered out of `item`

### feedback on:
- would this affect how the external service that's consuming from the data loader? as we are removing ids that's not present in our db

Ticket: https://getpocket.atlassian.net/browse/INFRA-803

### DIFF
--- a/src/dataLoader/savedItemsDataLoader.ts
+++ b/src/dataLoader/savedItemsDataLoader.ts
@@ -3,7 +3,6 @@ import { SavedItemDataService } from '../dataService';
 import { IContext } from '../server/context';
 import { SavedItem } from '../types';
 import { reorderResultByKey } from './utils';
-import * as Sentry from '@sentry/node';
 
 /**
  * Batch loader function to get saved items by URLs
@@ -18,10 +17,12 @@ export async function batchGetSavedItemsByUrls(
     context
   ).batchGetSavedItemsByGivenUrls(urls);
 
-  return reorderResultByKey<SavedItem, 'url'>(
+  const items = reorderResultByKey<SavedItem, 'url'>(
     { key: 'url', values: urls },
     savedItems
   );
+
+  return items.filter((item) => item != undefined);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/dataLoader/utils.spec.ts
```diff
@@ -25,6 +25,24 @@ describe('reorder by key', () => {
...
+        name: '1',
+        next: { id: 1, name: '2', next: null },
+      },
+      {
+        id: 2,
+        name: '2',
+        next: null,
+      },
+    ];
+    const expected = [undefined, results[0]];
```
sri-kirsh, line 41: utility function does the mapping for the input given to it..decided not to filter out `undefined` here as it can be handled by calling function if required.

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -29,21 +34,25 @@ export async function batchGetSavedItemsByUrls(
...
 
-  return reorderResultByKey<SavedItem, 'id'>(
+  const items = reorderResultByKey<SavedItem, 'id'>(
     { key: 'id', values: ids },
     savedItems
   );
+
+  //coz we filtered deleted items from savedItems,
+  //and may be items for the given Id is missing.
+  //so this function could map ids with undefined
```
sri-kirsh, line 54:
```suggestion
  //so reorderResultByKey function could map ids with undefined
```

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -10,17 +10,22 @@ import { reorderResultByKey } from './utils';
...
 ): Promise<SavedItem[]> {
-  const savedItems = await new SavedItemDataService(
-    context
-  ).batchGetSavedItemsByGivenUrls(urls);
+  let savedItems: SavedItem[] =
+    await savedItemService.batchGetSavedItemsByGivenUrls(urls);
 
-  return reorderResultByKey<SavedItem, 'url'>(
+  savedItems = savedItems.filter((item) => {
+    return item.status != 'DELETED';
```
kschelonka, line 20: aha

kschelonka (approved):

PR #568, opened 2022-11-18
fix(parser): retrying parser api calls

bassrock:
## Goal

Attempt to reduce https://sentry.io/organizations/pocket/issues/3439780078/?project=5761820&query=is%3Aunresolved&referrer=issue-stream by adding exponential backoff to the parser calls.

Upon looking at the breadcrumbs, items that are hitting the failing, work when I try them again manually indicating a transient error with the parser.

This is a stop gap, long term we need to look at the parser to understand why some API calls fail in this manner.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       - TZ=US/Central
 
   localstack:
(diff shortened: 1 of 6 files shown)


bassrock:
Talked with @jpetto decided to go with #569 as it fits our needs. Some reason the backoff library has issues with apollo and will not execute more then 1 retry

PR #569, opened 2022-11-18
fix(retry): adding very simple retry logic to the parser

bassrock:
## Goal
Alternative to https://github.com/Pocket/list-api/pull/568

Attempt to reduce https://sentry.io/organizations/pocket/issues/3439780078/?project=5761820&query=is%3Aunresolved&referrer=issue-stream by adding exponential backoff to the parser calls.

Upon looking at the breadcrumbs, items that are hitting the failing, work when I try them again manually indicating a transient error with the parser.

This is a stop gap, long term we need to look at the parser to understand why some API calls fail in this manner.

### DIFF
--- a/src/externalCaller/parserCaller.ts
+++ b/src/externalCaller/parserCaller.ts
@@ -1,4 +1,4 @@
-import fetch from 'node-fetch-retry';
+import fetch from 'node-fetch';
 import config from '../config';
 
 export type ItemResponse = {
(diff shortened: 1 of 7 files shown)


src/externalCaller/parserCaller.ts
```diff
@@ -33,4 +35,21 @@ export class ParserCaller {
       title: item.title ?? '',
     };
   }
+
+  public static async getOrCreateItem(
+    url: string,
+    tries = 3
```
jpetto, line 41: should this be in a config? i do not feel strongly - just asking.

jpetto (dismissed):

jpetto (approved):

PR #573, opened 2022-11-23
fix(data loaders): for non-existent items, set dataloader response as undefined but don't pass them to the url dataloader. replicating same behavior for url datalaoder. 

sri-kirsh:
## Goal
if we remove the Ids/url that is deleted/non-existent from the dataloaders response, it finds a mismatch between the ids passed in the request and the response it receives and throws [this  error](https://sentry.io/organizations/pocket/issues/3747671726/?project=5761820&referrer=project-issue-stream)

so to suppress the[ earlier undefined erro](https://sentry.io/organizations/pocket/issues/3081019079/?project=5761820&query=is%3Aunresolved&referrer=issue-stream)r, we just don't pass it to the alternate dataloader when item is not found for the given id/url, but return them key-undefined value in the response.

added test to make sure the dataloader call doesn't fail, and modified existing unit test to check all ids are returned with either savedItem or undefined value

ticket: https://getpocket.atlassian.net/browse/INFRA-803

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -417,12 +417,11 @@ class ListAPI extends TerraformStack {
         targetMaxCapacity: 10,
       },
       alarms: {
-        //TODO: When we start using this more we will change from non-critical to critical
         http5xxErrorPercentage: {
           threshold: 25,
           evaluationPeriods: 4,
           period: 300,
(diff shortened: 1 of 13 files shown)


kschelonka (approved):

sri-kirsh:
will merge next week as we are on code freeze

PR #574, opened 2022-11-23
bugfix(tags): Fix empty tag ID issue [INFRA-802]

kschelonka:
## Goal
Fix the issue with empty tag strings causing empty tag ids and then wreaking havoc throughout our application.

- Consolidate input validation/sanitization and other business logic (ID generation) into a tag model class, separate from the data service layer and dataloaders
- Point resolvers through the tag model rather than directly to the data service

- Keeping event emitter logic in the mutation resolver (convenience)

## Implementation Decisions
- Remove createTags mutation since it's duplicated by others and unused
- Didn't include every tag-related resolver in the model (just the ones affected by this bugfix), to reduce the already big scope
## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-802

Documentation:
* Project doc


src/dataService/tagDataService.ts
```diff
@@ -140,35 +132,21 @@ export class TagDataService {
...
-  public async getTagByName(tagName: string): Promise<Tag> {
-    const result = await this.getTagsByUserSubQuery().where(
-      'tag',
-      cleanAndValidateTag(tagName)
-    );
-    return TagObjectMapper.mapDbModelToDomainEntity(result[0]);
+    return tags.map(TagModel.toGraphqlEntity);
   }
 
-  public async getTagById(tagId: string): Promise<Tag> {
```
kschelonka, line 162: Removed because there is no such thing as a tag id in the database, so it shouldn't be in the data service

src/dataService/tagDataService.ts
```diff
@@ -199,35 +177,11 @@ export class TagDataService {
...
   }
 
-  /**
-   * Deduplicate a batch of tags prior to inserting in the
-   * database. Compares values for all keys of the TagCreateInput type.
-   * The keys aren't available until compile time, but if they get changed
-   * the linter should remind the dev to update.
-   * Best if this method is run after tags are 'cleaned'.
-   */
-  public static deduplicateTagInput(
```
kschelonka, line 214: Moved to tag model

src/dataService/utils.spec.ts
```diff
@@ -1,48 +1,5 @@
-import { strings } from 'locutus/php/';
 import { expect } from 'chai';
-import {
-  cleanAndValidateTag,
-  decodeBase64ToPlainText,
-  mysqlTimeString,
-} from './utils';
-
-describe('cleanAndValidateTag', () => {
```
kschelonka, line 9: this got moved to the tag model spec test

src/dataService/utils.ts
```diff
@@ -1,37 +1,5 @@
...
- *  3. Replace the unicode replacement character with ?, if present
- *  4. Truncate to 25 characters (an emoji counts as 1 character even if
- *     represented with multiple code points)
- *  5. Apply php addslashes function (ported to ts)
- *  6. Validates that the tag string is not empty, else throws an error
- * @param tagName the raw tag string
- * @returns string: the cleaned tag
- * @throws Error if cleaning results in an empty string
- */
-export function cleanAndValidateTag(tagName: string): string {
```
kschelonka, line 18: moved to tag model

src/models/tag.spec.ts
```diff
@@ -0,0 +1,90 @@
+import chai, { expect } from 'chai';
+import { TagCreateInput } from '../types';
+import deepEqualInAnyOrder from 'deep-equal-in-any-order';
+import * as tagModel from './tag';
+import { strings } from 'locutus/php';
+
+chai.use(deepEqualInAnyOrder);
+describe('tag model', () => {
+  describe('id', () => {
```
kschelonka, line 9: this describe block is the only new test in the file (others were moved)

src/dataService/tagDataService.ts
```diff
@@ -140,35 +132,21 @@ export class TagDataService {
       .whereIn('tag', latestTags)
       .orderBy('_updatedAt', 'desc');
 
-    return tags.map(TagObjectMapper.mapDbModelToDomainEntity);
+    return tags.map(TagModel.toGraphqlEntity);
   }
 
   public async getTagsByName(names: string[]): Promise<Tag[]> {
```
sri-kirsh, line 138: just checking if this method should call sanitizeTag as the old method does clean&validateTag

src/models/tag.ts
```diff
@@ -0,0 +1,360 @@
...
+  /**
+   * Fetch multiple Tags by ID
+   * Will not throw a NotFoundError if a tag ID is missing,
+   * so the size of the response might differ from the number
+   * of IDs requested.
+   * @param ids array of tag IDs to retrieve
+   * @returns An array of Tag entities
+   */
+  public getManyById(ids: string[]): Promise<Tag[]> {
+    // TODO: Not found partial repsonse?
```
sri-kirsh, line 149: i couldn't find the usage of this function..but assuming this is used in bulk calls, if one of them is invalid, we decided to fail the whole call..so we don't need to handle partial response

src/test/functional/queryServices.test/tags.integration.ts
```diff
@@ -317,60 +387,17 @@ describe('tags query tests - happy path', () => {
...
     });
     expect(res.data?._entities[0].savedItemById.url).to.equal('http://abc');
     expect(res.data?._entities[0].savedItemById.tags[0].name).to.equal(
       'travel'
     );
-    const tagId0 = Buffer.from(
-      res.data?._entities[0].savedItemById.tags[0].id,
-      'base64'
-    ).toString();
+    const tagId0 = TagModel.decodeId(
```
sri-kirsh, line 398: not a blocker: thinking if we should use the same resource for code and test?

src/dataService/tagDataService.ts
kschelonka: no, since this is coming from the database any tags should already be sanitized

src/models/tag.ts
kschelonka: I forgot to add this to the resolvers it looks like. thank you!

src/test/functional/queryServices.test/tags.integration.ts
kschelonka: yeah, in this case I was taking a shortcut so I didn't have to update every test with the base64 strings 😬 

i'll update

src/models/tag.ts
```diff
@@ -0,0 +1,360 @@
...
+   * @returns the encoded ID
+   */
+  public static encodeId(name: string): string {
+    return Buffer.from(name + config.data.tagIdSuffix).toString('base64');
+  }
+
+  /**
+   * Associate one or more tags to a save
+   */
+  public async createTagSaveConnections(
```
sri-kirsh, line 78: trying to understand this workflow:

goes like resolver (calling data models and event emission)-> dataModel (validation, handle business logic and db calls) -> dataService(for db read/write)?

kschelonka: yes, but I think the resolver could just be a shallow passthrough in the future and event emission handled in the model? what do you think?

src/dataService/tagDataService.ts
sri-kirsh: just trying to understand our workflow here:

the `validateTag` function is being used by read calls as we call `toGraphqlEntity`

so in read, we only check if the required fields are set and we allow any invalid input to passthrough (e.g empty tag name as they are valid in our db as of now)

 whereas in write, we validate the tags format (e.g empty string/character limit etc) and block any invalid input (e.g empty tag string)

kschelonka: Yep! you got it. The `toGraphQlEntity` validation I'm seeing as more of a developer check to make sure that we have included all the required fields in the data layer response, since we don't have type safety and also the db model does not match our graphql entity

src/models/tag.ts
sri-kirsh: ya just like db write, we might have to do the same tag-sanitization before we emit events. in future, we can consolidate them

sri-kirsh (approved):

src/models/tag.ts
kschelonka: I'll make a ticket so we don't forget. I just didn't want to shove too much into this already big pr 😓

PR #586, opened 2022-12-02
fix(item): resolve savedItem on Item [INFRA-778]

kschelonka:
# Goal

Fixes bug where itemByItemId and itemByItemUrl
queries attempting to fetch savedItem field
on Item always returned null

Conceptually this is because we're resolving an `Item` field (`Item.savedItem`) on the list subgraph, but we didn't actually define a reference resolver. The key for the entity is available via federation, but it needs the reference resolver in order to get passed down to the rest of the resolver chain.

## Implementation decisions
- Replaced old data service call with dataloader in `Item.savedItem` resolver

## References

Jira ticket:
* [INFRA-778]

### DIFF
--- a/src/resolvers/index.ts
+++ b/src/resolvers/index.ts
@@ -22,9 +22,10 @@ import {
   upsertSavedItem,
 } from './mutation';
 import { tagsSavedItems } from './tag';
-import { SavedItem, Tag } from '../types';
+import { Item, SavedItem, Tag } from '../types';
 import { IContext } from '../server/context';
 import { writeClient } from '../database/client';
+import { NotFoundError } from '@pocket-tools/apollo-utils';
 
 const resolvers = {
   ItemResult: {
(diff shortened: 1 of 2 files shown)


src/resolvers/index.ts
```diff
@@ -38,7 +39,23 @@ const resolvers = {
     tags: userTags,
   },
   Item: {
-    savedItem,
+    savedItem: async (item: Item, args, context: IContext) => {
+      const save = await context.dataLoaders.savedItemsByUrl.load(
+        item.givenUrl
+      );
+      if (save == null) {
+        throw new NotFoundError(`No Save found for url=${item.givenUrl}`);
```
kschelonka, line 47: Should we return NotFoundError here or just null?

src/dataLoader/savedItemsDataLoader.ts
```diff
@@ -13,13 +13,9 @@ export async function batchGetSavedItemsByUrls(
...
   urls: string[]
 ): Promise<SavedItem[]> {
-  let savedItems: SavedItem[] =
+  const savedItems: SavedItem[] =
     await savedItemService.batchGetSavedItemsByGivenUrls(urls);
 
-  savedItems = savedItems.filter((item) => {
-    return item.status != 'DELETED';
-  });
-
```
jpetto, line 22: just checking - will this impact non-mobile clients in a negative way? not sure how widely this change needs to be checked/circulated.

jpetto (approved):
code-wise looks great. only questions are expectation related - but don't need to block if those concerns have already been addressed.

src/resolvers/index.ts
kschelonka: It's kind of inconsistent. We do have a contract to return NotFoundError for specific id queries e.g. `savedItemById`. Good call to ask the clients... I would certainly hope that they would not just error out if there's data + errors haha

src/dataLoader/savedItemsDataLoader.ts
kschelonka: No, since web should be (and is) using filters to retrieve saves associated with specific views (e.g. My List, Archive)

src/resolvers/index.ts
kschelonka: https://pocket.slack.com/archives/C01VDFM30J2/p1670263011027919

This was uncovered during some QA testing and isn't currently included in client queries. I think the NOT_FOUND makes sense as additional data here, but we can revisit the contract if clients start to use it.

PR #599, opened 2022-12-15
feat(deps): Migrate to apollo v4 [INFRA-842]

kschelonka:
## Goal
Apollo V4 migration! I really wanted to get this in before the code freeze to see if the error handler updates fixed some of the bugs we saw, but oh well. 

There's one random test failing, but I really don't know what differentiates it from others...?

- Factor out the apollo server into its own file for import purposes
- Use full e2e testing with supertest to replace `server.executeOperation`; this was a really straightforward migration and in some ways was actually easier to pass in different headers, properly make the context, etc.
- Reorganize test folders to separate graphql vs. other integration tests
- Remove the test server and just use the prod app 

- Why do you think this one test is failing? It seems like the same as all the other tests but is getting an ECONN:REFUSED error...

## Implementation Decisions
- Mostly the "extra" things I did were reorganizing the test folders -- I think this makes more sense? but let me know (might want to look at the file tree to see).
- e2e test to replace `server.executeOperation`

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-842


src/server/apollo.ts
```diff
@@ -0,0 +1,93 @@
...
+  unifiedEventHandler,
+} from '../businessEvents';
+import queueDeleteRouter from './routes/queueDelete';
+import { BatchDeleteHandler } from '../aws/batchDeleteHandler';
+import { EventEmitter } from 'events';
+import { initAccountDeletionCompleteEvents } from '../aws/eventTypes';
+import { typeDefs } from './typeDefs';
+import { resolvers } from '../resolvers';
+
+export async function startServer(port: number) {
```
hyperparabolic, line 28: Nice, I was thinking almost the exact same thing here. Also TIL port 0 dynamically allocates a port.

Do we want to make this accept an entire config object as a parameter and pass it through in main? We'd be just a few steps away from being able to run tests in parallel.

src/models/tag.ts
```diff
@@ -300,7 +298,7 @@ const validateTag = (tag: any): true => {
 
   if (err) {
     Sentry.captureException(err);
-    throw new ApolloError(err);
+    throw new Error(err);
```
hyperparabolic, line 301: I think this might be getting captured twice with the sentry plugin. I think we might also want this to be a `GraphQLError` to match the old behavior. Plain JS errors get masked to a InternalServerError.

src/resolvers/mutation.ts
```diff
@@ -55,7 +54,7 @@ export async function upsertSavedItem(
 
     if (upsertedItem == undefined) {
       console.info(`savedUrl: ${savedItemUpsertInput.url}`);
-      throw new ApolloError(`unable to add an item`);
+      throw new Error(`unable to add an item`);
```
hyperparabolic, line 57: Same question, do we want this to be a GraphQLError?

src/resolvers/mutation.ts
```diff
@@ -74,9 +73,7 @@ export async function upsertSavedItem(
   } catch (e) {
     console.log(e.message);
     Sentry.captureException(e);
-    throw new ApolloError(
-      `unable to add item with url: ${savedItemUpsertInput.url}`
-    );
+    throw new Error(`unable to add item with url: ${savedItemUpsertInput.url}`);
```
hyperparabolic, line 76: Ditto, `GraphQLError`?

src/server/apollo.ts
```diff
@@ -0,0 +1,93 @@
...
+  const app = express();
+  // Our httpServer handles incoming requests to our Express app.
+  // Below, we tell Apollo Server to "drain" this httpServer,
+  // enabling our servers to shut down gracefully.
+  const httpServer = http.createServer(app);
+  // Initialize routes
+  app.use('/queueDelete', queueDeleteRouter);
+
+  // Start BatchDelete queue polling if not test env
+  if (process.env.NODE_ENV != 'test') {
```
hyperparabolic, line 43: Nit, grab this from `config.app.environment`, especially if we start passing it as a parameter.

src/server/apollo.ts
```diff
@@ -0,0 +1,93 @@
...
+      dbClient: readClient(),
+      eventEmitter: itemsEventEmitter,
+    });
+  };
+
+  const server = new ApolloServer<ContextManager>({
+    schema: buildSubgraphSchema({ typeDefs, resolvers }),
+    plugins: [
+      sentryPlugin,
+      process.env.NODE_ENV === 'production'
```
hyperparabolic, line 68: Nit, grab this from `config.app.environment`, especially if we start passing it as a parameter.

src/server/apollo.ts
```diff
@@ -0,0 +1,93 @@
...
+      request: req,
+      dbClient: readClient(),
+      eventEmitter: itemsEventEmitter,
+    });
+  };
+
+  const server = new ApolloServer<ContextManager>({
+    schema: buildSubgraphSchema({ typeDefs, resolvers }),
+    plugins: [
+      sentryPlugin,
```
hyperparabolic, line 67: It's a pretty small change to incorporate `ApolloServerPluginDrainHttpServer` here, and get graceful shutdowns.  Do we want to do that?  More info at this link: https://www.apollographql.com/docs/apollo-server/migration/#warning-for-servers-without-draining

src/server/apollo.ts
```diff
@@ -0,0 +1,93 @@
...
+      process.env.NODE_ENV === 'production'
+        ? ApolloServerPluginLandingPageDisabled()
+        : ApolloServerPluginLandingPageGraphQLPlayground(),
+      ApolloServerPluginDrainHttpServer({ httpServer }),
+    ],
+    formatError: process.env.NODE_ENV !== 'test' ? errorHandler : undefined,
+    introspection: process.env.NODE_ENV !== 'production',
```
hyperparabolic, line 74: Nit, grab this from config.app.environment, especially if we start passing it as a parameter.

hyperparabolic (changes requested):
I plan to re-review tomorrow and see if I can help debug that test, but here's a start with a few questions / suggestions.

src/server/apollo.ts
kschelonka: Sure yeah it's a possibility, but it would only work for any tests that didn't use the shared db backend (so query tests only). I don't really have a strong opinion.

src/server/apollo.ts
kschelonka: sure, although I will probably say we should follow up in a separate ticket since we are using node env elsewhere and I was trying to limit unrelated changes

src/server/apollo.ts
kschelonka: It's in line 71 :)

src/server/apollo.ts
kschelonka: see above, limiting unrelated changes (this file was just renamed basically)

src/resolvers/mutation.ts
kschelonka: Can put this in a follow-up ticket as I was trying to maintain current behavior as much as possible and limit unrelated changes.

src/models/tag.ts
kschelonka: Seeing some similar threads here, will put it on my list to double check the behavior that's intended.

src/server/apollo.ts
hyperparabolic: It's definitely out of scope for this PR, but the SQL database names and dynamodb table names are in the config file. If we randomly generated those, migrated, seeded, and then cleaned up as a part of the test lifecycle it would be possible here as well.  Then you could probably run the integration tests in watch mode as well while developing.

hyperparabolic (approved):
Alright, I'm cool with everything commented here being moved to another PR. I don't see anything else. Approved.

src/server/apollo.ts
kschelonka: Yeah it's a cool idea and I think worthwhile to consider for future repos, particularly ones that actually are sagas and have control of the db. Just thinking of cost-benefit for this particular repo

PR #601, opened 2022-12-19
feat(PocketSave): Create a PocketSave entity that is retrievable by ID for now

cmharlow:
## Goal

This PR adds the first pass as a PocketSave entity, which is scheduled to eventually replace SavedItem.

To make the split easier, PocketSave has largely used code separated from SavedItem.

It's also been setup to follow a proposed new pattern where business logic lives in models, data source / database logic lives in dataservice, and resolvers are as dumb as possible.

- getting date handling for 0000-00-00 00:00:00 was a real pain in the ass, since the value coming from the knex response sometimes would be a Date object returning NaN, other times it would be a string. Probably should move the existing checks into utils, but would appreciate other eyes & thoughts
- this is a new setup, would appreciate feedback on it

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-821

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,6 +1,9 @@
+"""
+ISOString scalar - all datetimes fields are Typescript Date objects on this server &
+returned as ISO-8601 encoded date strings (e.g. ISOString scalars) to GraphQL clients.
+See Section 5.6 of the RFC 3339 profile of the ISO 8601 standard: https://www.ietf.org/rfc/rfc3339.txt.
+"""
 scalar ISOString
-
-scalar Markdown
 
 scalar Url
 
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,3 +1,7 @@
+scalar ISOString
+
+scalar Markdown
+
 scalar Url
 
 """
@@ -495,6 +499,11 @@ extend type User @key(fields: "id") {
   Get a SavedItem by its id
   """
   savedItemById(id: ID!): SavedItem
+
+  """
+  Get a PocketSave by its id
+  """
+  pocketSaveById(id: ID!): PocketSave
 }
 
 """
@@ -620,3 +629,145 @@ type Mutation {
   """
   replaceSavedItemTags(input: [SavedItemTagsInput!]!): [SavedItem!]!
 }
+
+# """
+
+# PocketSave Entities (a rework of SavedItem) starts here.
+
+# """
+
+"""
+Enum to specify the PocketSave Status (mapped to integers in data store).
+"""
+enum PocketSaveStatus {
+  ARCHIVED
+  DELETED
+  HIDDEN
+  UNREAD
+}
+
+"""
+New Pocket Save Type, replacing SavedItem.
+
+Represents a Pocket Item that a user has saved to their list.
+(Said otherways, indicates a saved url to a users list and associated user specific information.)
+"""
+type PocketSave {
+  # Not resolved on this subgraph.
+  # """
+  # Authors credited for content.
+  # """
+  # authors: [Authors!]
+
+  """
+  Indicates if the PocketSave is archived.
+  """
+  archived: Boolean!
+
+  """
+  Timestamp that the PocketSave became archived, null if not archived.
+  """
+  archivedAt: ISOString
+
+  """
+  Unix timestamp of when the PocketSave was created.
+  """
+  createdAt: ISOString!
+
+  """
+  Unix timestamp of when the entity was deleted.
+  """
+  deletedAt: ISOString
+
+  # Not resolved on this subgraph.
+  # """
+  # Excerpt from text content, typically the first paragraph.
+  # """
+  # excerpt: Markdown
+
+  """
+  Indicates if the PocketSave is favorited.
+  """
+  favorite: Boolean!
+
+  """
+  Timestamp that the PocketSave became favorited, null if not favorited.
+  """
+  favoritedAt: ISOString
+
+  """
+  The url the user gave (as opposed to normalized URLs).
+  """
+  givenUrl: String!
+
+  """
+  Surrogate primary key.
+  """
+  id: ID!
+
+  # Not resolved on this subgraph.
+  # """
+  # Reference to the underlying Resource that this PocketSave metadata is about.
+  # """
+  # resource: PocketResource!
+
+  # Not resolved on this subgraph.
+  # """
+  # The web site publisher; defaults to the domain of the PocketSave URL.
+  # """
+  # source: String!
+
+  """
+  The status of this PocketSave; Marked for review for possible removal.
+  """
+  status: PocketSaveStatus
+
+  """
+  The Suggested Tags associated with this PocketSave, if the user is not premium or there are none, this will be empty.
+  """
+  suggestedTags: [Tag!]
+
+  """
+  The Tags associated with this PocketSave.
+  """
+  tags: [Tag!]
+
+  # Not resolved on this subgraph.
+  # """
+  # Direction text should be displayed, e.g. right-to-left; an enum.
+  # """
+  # textDirection: TextDirection
+
+  # Not resolved on this subgraph.
+  # """
+  # Image to display as thumbnail.
+  # """
+  # thumbnail: Image
+
+  # Not resolved on this subgraph.
+  # """
+  # Estimated time to consume content, rounded to the nearest minute.
+  # """
+  # timeToConsume: Int
+
+  """
+  The title of the Resource; defaults to the URL.
+  """
+  title: String!
+
+  """
+  Unix timestamp of when the PocketSave was last updated, if any property on the PocketSave is modified this timestamp is set to the modified time.
+  """
+  updatedAt: ISOString
+}
+
+# To be setup: connect PocketSaves to a new entity - PocketResource
+# extend type PocketResource @key(fields: "itemId") {
+#   "key field to identify the PocketResource entity in the Parser service"
+#   itemId: ID! @external
+
+#   """
+#   Identify if the given PocketResource is in the user's list.
+#   """
+#   pocketSave: PocketSave
+# }
--- a/src/dataService/pocketSavesService.ts
+++ b/src/dataService/pocketSavesService.ts
@@ -1,0 +1,155 @@
+import { Knex } from 'knex';
+import { IContext } from '../server/context';
+import { PocketSaveStatus } from '../types';
+
+export type RawListResult = {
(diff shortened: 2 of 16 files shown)


src/test/graphql/queries/pocketSave.integration.ts
```diff
@@ -0,0 +1,183 @@
...
+  const date1 = new Date('2008-10-21 13:57:01');
+  const date2 = new Date('0000-00-00 00:00:00');
+  const date3 = new Date('2008-10-21 14:00:01');
+  const date4 = new Date('2012-08-13 15:32:05');
+  const date5 = new Date('2008-11-03 08:51:01');
+  let app: Express;
+  let server: ApolloServer<ContextManager>;
+  let url: string;
+
+  const GET_POCKET_SAVE = `
```
hyperparabolic, line 20: I know this isn't happening in every repo, but I would recommend making this a `gql` tag and using the graphql print function to get these converted back to strings. Imports for these if you haven't seen them:

```
import { gql } from 'graphql-tag';
import { print } from 'graphql';
```

It's nice for us just semantically to be able to find all of our queries in a repo by searching for a string, it's nice to get syntax highlighting, and we're likely going to need them if we ever want to introduce graphql query linting without it being a huge pain in the butt later.

src/models/tag.spec.ts
```diff
@@ -6,6 +6,8 @@ import { strings } from 'locutus/php';
 
 chai.use(deepEqualInAnyOrder);
 describe('tag model', () => {
+  // write test for getSuggestedTags checking isPremium conditional
```
hyperparabolic, line 9: Is this a TODO for this PR?

src/dataService/pocketSavesService.ts
```diff
@@ -0,0 +1,155 @@
...
+  public static convertListResult(listResult: RawListResult[]): ListResult[];
+  /**
+   * Convert the `status` field in the list table to the expected
+   * GraphQL ENUM string
+   * @param listResult
+   */
+  public static convertListResult(
+    listResult: RawListResult | RawListResult[] | null
+  ): ListResult | ListResult[] | null {
+    if (listResult == null) {
```
hyperparabolic, line 69: I would probably go for `listResult === null` if you're interested in null in particular, or `!listResult` if you're interested in any falsy value.

src/models/pocketSave.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * * Fetch a PocketSave by its ID
+   * * @param id the ID of the PocketSave to retrieve
+   * @throws NotFoundError if the PocketSave does not exist
+   * @returns the PocketSave entity
+   */
+  public async getById(id: string): Promise<PocketSave> {
+    const listRow = await this.saveService.getListRowById(id);
+    if (listRow == null) {
```
hyperparabolic, line 42: Ditto here.  `listRow === null` or `!listRow`.

hyperparabolic (changes requested):
Good work! Just a few questions, suggestions, and requests.

src/dataService/pocketSavesService.ts
kschelonka: we use this pattern when the response could be null or undefined, but we don't care which. I like that it's a little more specific than just falsy values, which can be a little unintuitive in js/ts (e.g. [] is truthy, {} is truthy)

src/server/context.ts
```diff
@@ -55,9 +57,22 @@ export class ContextManager implements IContext {
...
+      pocketSave: new PocketSaveModel(this),
     };
   }
-  models: { tag: TagModel };
+  models: {
+    tag: TagModel;
+    pocketSave: PocketSaveModel;
+  };
+
+  withDbClientOverride(dbClient: Knex): ContextManager {
```
sri-kirsh, line 68: with our new change on choosing dbClient at ContextManager based on request, we don't need this function right?

src/test/graphql/queries/pocketSave.integration.ts
sri-kirsh: agree, that's the pattern we follow in other repo, e.g here: https://github.com/Pocket/user-api/blob/1b63de9feafa18772d6f74cdc27426a2b384f819/src/test/functional/context.integration.ts#L46

src/models/pocketSave.ts
```diff
@@ -0,0 +1,48 @@
+import { PocketSave } from '../types';
+import { IContext } from '../server/context';
+import { ListResult, PocketSaveDataService } from '../dataService';
+import { NotFoundError } from '@pocket-tools/apollo-utils';
+
+export class PocketSaveModel {
```
sri-kirsh, line 6: nit: class docs - looking for some details on responsibility of this class with mutation and query workflow

sri-kirsh: so models will be our service layer, and `*Service` will be our dbAccessLayer

src/test/graphql/queries/pocketSave.integration.ts
```diff
@@ -0,0 +1,183 @@
...
+  it('should return null if no item is found for the user', async () => {
+    const variables = {
+      userId: '1',
+      itemId: '10',
+    };
+    const res = await request(app).post(url).set(headers).send({
+      query: GET_POCKET_SAVE,
+      variables,
+    });
+    expect(res.body.data?._entities[0].pocketSaveById).toBe(null);
```
sri-kirsh, line 138: do we also want to assert on error code is NOT_FOUND?

src/server/context.ts
cmharlow: @sri-kirsh this is mess i picked up from your changes I believe. I'll repair it shortly.

src/models/pocketSave.ts
cmharlow: Ask @kschelonka but yes that is my understanding and what I tried to line out in the PR description.

sri-kirsh: yup, would be good to have it in class docs  :)

schema.graphql
```diff
@@ -1,3 +1,7 @@
+scalar ISOString
```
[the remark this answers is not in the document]
cmharlow: okay, tried to add something and I removed Markdown since its for a field we're likely not using yet here.

src/dataService/pocketSavesService.ts
cmharlow: Trying the more explicit `x === undefined || x === null`

src/dataService/pocketSavesService.ts
```diff
@@ -0,0 +1,155 @@
...
+      return null;
+    }
+
+    const statusConvert = (row: RawListResult) => {
+      console.log(typeof row.time_favorited);
```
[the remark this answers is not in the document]
cmharlow: dang yes it is, sorry i didn't catch that

src/dataService/pocketSavesService.ts
```diff
@@ -0,0 +1,155 @@
...
+   * @param listResult
+   */
+  public static convertListResult(
+    listResult: RawListResult | RawListResult[] | null
+  ): ListResult | ListResult[] | null {
+    if (listResult == null) {
+      return null;
+    }
+
+    const statusConvert = (row: RawListResult) => {
```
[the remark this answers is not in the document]
cmharlow: alright i'm trying that and i read a bit about closures, thanks for this

src/dataService/pocketSavesService.ts
```diff
@@ -0,0 +1,155 @@
...
+        api_id: row.api_id,
+        api_id_updated: row.api_id_updated,
+        favorite: row.favorite,
+        given_url: row.given_url,
+        item_id: row.item_id,
+        resolved_id: row.resolved_id,
+        status: PocketSaveDataService.statusMap[row.status],
+        time_added:
+          row.time_added instanceof Date
+            ? !isNaN(row.time_added.getTime())
```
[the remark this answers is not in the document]
cmharlow: yeah that makes sense. I had this on my to revisit list, i was just so pissed at that 0000-00-00 stuff last night haha. Rewriting now.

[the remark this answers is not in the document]
cmharlow: tried a rewrite, moved the new function to utils and added a few tests for it.

src/models/pocketSave.ts
cmharlow: updating with the same proposed change which makes it more explicit.

src/models/tag.spec.ts
cmharlow: shoot yeah. sorry i let that slide too. Adding that test now.

src/test/graphql/queries/pocketSave.integration.ts
cmharlow: makes sense. a lot of this is copypasta from the existing setup for pre-existing code, but going to try to rework now to follow that setup.

src/test/graphql/queries/pocketSave.integration.ts
cmharlow: this is just copypasta from pre-existing structure with some other expansions, so will add this too.

cmharlow: believe i added a check for error response and error code along with existing null body response.

schema.graphql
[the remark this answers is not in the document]
hyperparabolic: Looks great.

src/dataService/pocketSavesService.ts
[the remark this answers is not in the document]
hyperparabolic: I like having no console as a linter rule, and then just generally allowing any exception for exactly this reason 😄

src/models/tag.spec.ts
hyperparabolic: Where did this get added?  I'm not sure I see it.

hyperparabolic (commented):
I think just one last question and then I'm good to go on this.

src/models/tag.ts
```diff
@@ -139,6 +140,19 @@ export class TagModel {
     return this.tagService.getTagsByUserItem(id);
   }
 
+  /**
```
cmharlow, line 143: bookmark.

hyperparabolic (approved):

cmharlow:
alright @hyperparabolic added those missing tests - thanks for catching that! I think its okay to go? lol who knows

hyperparabolic (approved):
Looks great.  Approved!

PR #614, opened 2023-01-10
Revert "Code Freeze 2022 Thaw"

cmharlow:
Reverts Pocket/list-api#607

We need to update the PR to have a new healthcheck for the LB, otherwise deploy fails

PR #615, opened 2023-01-10
Revert "Revert "Code Freeze 2022 Thaw""

kschelonka:
Reverts Pocket/list-api#614


kschelonka:
Request changes: Need to fix the health check endpoint so it doesn't hang in codebuild

sri-kirsh (approved):

PR #627, opened 2023-02-01
fix(replaceSavedItemTags):  make getTagsBySavesId async and added test to validate we can re-add old tags

sri-kirsh:
## Goal
bug: if we replace tags for savedItems, the response contains the old tag values.

## investigation
the main branch:
- [x] works locally and we can re-add an existing tag, added test to verify this.
- [x] works by directly hitting the subgraph in dev environment 
- [ ] does not works by directly hitting the subgraph in prod environment 
          - extremely rare case in prod, with long or unusal strings (so random) updates to the correct call- suspecting it to be some async workflow execution :: read prioritized over write?

##implementation/feedback on:
- making getTagsbySaveId as async function as it returns `promise<tag[]>`. not sure if the `tags` resolver gets prioritized before the mutation is complete as mutation is async but getBySaveId is sync 

- [ ] test in prod after merging

Ticket: INFRA-951


src/models/tag.ts
```diff
@@ -137,7 +137,7 @@ export class TagModel {
     return tag;
   }
 
-  public getBySaveId(id: string): Promise<Tag[]> {
+  public async getBySaveId(id: string): Promise<Tag[]> {
```
sri-kirsh, line 140: thinking if this has to be `async` method as it returns promise.Tag and not having this somehow prioritizes resolving `tags` before completing the write

src/resolvers/savedItem.ts
```diff
@@ -13,7 +13,7 @@ export async function tags(
   args,
   context: IContext
 ): Promise<Tag[]> {
-  return context.models.tag.getBySaveId(parent.id);
+  return await context.models.tag.getBySaveId(parent.id);
```
sri-kirsh, line 16: added this for my own readability - my understanding is having `await` or no in this workflow should not affect its child resolvers

bassrock (approved):

PR #628, opened 2023-02-01
fix(dev env): trying to fix dev environment code build 

sri-kirsh:
## Goal
 trying to fix dev environment code build

code build successful: https://us-east-1.console.aws.amazon.com/codesuite/codebuild/410318598490/projects/ListAPI-Dev/history?region=us-east-1&builds-meta=eyJmIjp7InRleHQiOiIifSwicyI6e30sIm4iOjIwLCJpIjowfQ

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -47,7 +47,7 @@ phases:
       - npm ci
       - npm run build
       # synthesize the js into terraform json with the proper node environment
-      - 'if [ "$GIT_BRANCH" == "$DEV_BRANCH" ]; then NODE_ENV=development npm run synth; else npm run synth; fi'
+      - 'if [ "$GIT_BRANCH" = "$DEV_BRANCH" ]; then NODE_ENV=development npm run synth; else npm run synth; fi'
       - cd cdktf.out/stacks/list-api
       - rm -rf .terraform # remove sym-link to ../.terraform (if this sym-link exists, terraform doesn't apply correctly. hmmmm)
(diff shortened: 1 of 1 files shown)


buildspec.yml
```diff
@@ -43,11 +43,11 @@ phases:
       - echo "//npm.pkg.github.com/:_authToken=${GITHUB_ACCESS_TOKEN}" > ~/.npmrc
       - echo Setting environment variables
       - cd .aws
-      - npm install
```
cmharlow, line 46: why remove npm install here? this repo appears to use cdktf, which may need those packages installed for the synth & terraform to run

buildspec.yml
```diff
@@ -43,11 +43,11 @@ phases:
...
       - tfenv install
+      - npm ci
       - npm run build
       # synthesize the js into terraform json with the proper node environment
-      - 'if [ "$GIT_BRANCH" == "$DEV_BRANCH" ]; then NODE_ENV=development npm run synth; else npm run synth; fi'
```
cmharlow, line 50: shit, thanks for catching this, i missed a few across repos. damn dash vs bash vs ash inconsistencies.

buildspec.yml
sri-kirsh: usually in ci, we run the the package lock command right? i.e what's tested in local is pushed to package-lock and that's run in CI

cmharlow (approved):

PR #630, opened 2023-02-02
feat(corpusItem): adding in savedItem on corpus item

bassrock:
## Goal

Clients need to know if a Corpus Item has been saved to a user's saves for display purposes. (ie. should they display a saved button on recomendations)

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,3 +1,5 @@
+extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@external"])
+
 scalar Url
 
 """
@@ -80,6 +82,15 @@ input PaginationInput {
   If `last` contains a value, `first` should be null/omitted in the input.
   """
   last: Int
+}
+
+type CorpusItem @key(fields: "url") {
+  url: Url! @external
+
+  """
+  The user's saved item, from the Corpus Item, if the corpus item was saved to the user's saves
+  """
+  savedItem: SavedItem
 }
 
 """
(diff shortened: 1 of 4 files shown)


schema.graphql
```diff
@@ -1,3 +1,5 @@
+extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@external"])
```
bassrock, line 1: switching to fed 2.0 for cleaner syntax

schema.graphql
```diff
@@ -460,7 +471,7 @@ type SavedItemTagAssociation {
   tagId: ID!
 }
 
-extend type User @key(fields: "id") {
+type User @key(fields: "id") {
```
bassrock, line 474: fed 2 drops extend syntax

src/resolvers/index.ts
```diff
@@ -39,13 +39,7 @@ const resolvers = {
   },
   Item: {
     savedItem: async (item: Item, args, context: IContext) => {
-      const save = await context.dataLoaders.savedItemsByUrl.load(
-        item.givenUrl
-      );
-      if (save == null) {
-        throw new NotFoundError(`No Save found for url=${item.givenUrl}`);
```
bassrock, line 46: we shouldnt be throwing a not found error here. we should just return null, and continue on. There are plent of cases where an Item won't return a savedItem and its not an error

kschelonka (dismissed):

src/test/graphql/queries/item.integration.ts
```diff
@@ -173,7 +173,7 @@ describe('item', () => {
     expect(entities.length).toEqual(1);
     expect(entities[0]).toEqual(expected);
   });
-  it('returns null with not found error if the save does not exist', async () => {
+  it('returns null if the save does not exist', async () => {
```
kschelonka, line 176: serious deja vu here, I thought we already did this :| but yes I agree

PR #631, opened 2023-02-03
fix(context): set context db client only once when request initialized

kschelonka:
## bug:
- child resolvers of a mutation resolver chain was using the readClient - [debug PR is here](https://github.com/Pocket/list-api/pull/629)

## fix:
- Regex the request to determine if a query or mutation; use write client if mutation, read otherwise.
- unit and integration test
- removed existing IContext method to amend the context for writeClient and resolver check to amend context for mutations

## References
See slack thread https://pocket.slack.com/archives/C03QVL4SU/p1675453150101759 

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-951


src/server/apollo.ts
```diff
@@ -27,6 +27,27 @@ import { EventEmitter } from 'events';
 import { initAccountDeletionCompleteEvents } from '../aws/eventTypes';
 import { typeDefs } from './typeDefs';
 import { resolvers } from '../resolvers';
+import { Knex } from 'knex';
+
+/**
```
sri-kirsh, line 32: other options considered is checking the resolver chain's `info` object on if the parent's Operation type is mutation..but this fix is more concrete at one place when the request comes in, and we don't have to amend context on every resolver

sri-kirsh (approved):

sri-kirsh:
LGTM, merging to test in prod

PR #637, opened 2023-02-10
feat(validation): Prototyping input validation with @constraint

kschelonka:
## Goal
Literally as I was looking at [graphql-constraint-directive](https://github.com/confuser/graphql-constraint-directive) and remembering why I couldn't test it out before, the maintainer merged a fix for it to work with apollo v4. Is it fate?

Note: I contributed a small typescript fix to the library which has been merged but not yet released -- that's what those TODOs are about. I'll also need to do a little more finangling to get this to build in the studio, but it works locally.

I think this is a nice way to go. It's self-documenting and testable. The only thing that we have to consider is that it can't be applied on the arguments, so we'll have to add input types for all the bulk mutations. This would technically be a change to the contract but I don't think it would be problematic. I think the only other viable alternative would be putting it directly in the resolver/model code, as writing our own plugin would be an unnecessary maintenance burden when graphql-constraint-directive already exists.

e.g. instead of 
```
tagRename(input: [TagRenameInput!]!, timestamp: ISOString!): TagWriteMutationPayload!
```

we'd need something like

```
input BulkTagRenameInput {
  inputs: [TagRenameInput!]! @constraint(minItems: 1, maxItems: 30)
  timestamp: ISOString!
}

tagRename(input: BulkTagRenameInput): TagWriteMutationPayload!
```

[INFRA-549]

[INFRA-549]: https://getpocket.atlassian.net/browse/INFRA-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -1,6 +1,6 @@
 extend schema
   @link(url: "https://specs.apollo.dev/link/v1.0")
-  @link(url: "https://specs.apollo.dev/federation/v2.3",
+  @link(url: "https://specs.apollo.dev/federation/v2.1",
         import: ["@key", "@composeDirective"])
   # The link directive is required to federate @constraint
   # It doesn't actually have to be a real spec, but it would be good
--- a/schema.graphql
+++ b/schema.graphql
@@ -2,6 +2,10 @@ extend schema
   @link(url: "https://specs.apollo.dev/link/v1.0")
   @link(url: "https://specs.apollo.dev/federation/v2.0",
         import: ["@key"])
+  # The link directive is required to federate @constraint
+  # It doesn't actually have to be a real spec, but it would be good
+  # to write one and replace this.
+  @link(url: "https://myspecs.dev/myDirective/v1.0", import: ["@constraint"])
   @composeDirective(name:"@constraint")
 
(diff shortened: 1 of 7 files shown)


schema.graphql
```diff
@@ -1,3 +1,5 @@
+extend schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@key", "@shareable"])
+
```
Herraj, line 2: I'm assuming we need this to use `@constraint` directive?

Herraj (approved):
Exciting!

schema.graphql
kschelonka: It was in the example documentation, so I was wondering if it was a fed 2 thing... but no, I don't think so... I'm still getting composition errors. I'll futz around with the schema tomorrow and see if I can get it to compose.

kschelonka: I think actually I have to follow this: https://www.apollographql.com/docs/federation/federated-types/federated-directives/#managing-custom-directives

kschelonka:
Implemented in #717

PR #643, opened 2023-02-16
bug(xray): replace node-autoinstr w/specific pkgs to unblock apollo-utils update

cmharlow:
## Goal

Dealing with more Xray & Opentelemetry setup bugs that are blocking #641 

Namely:
* @opentelemetry/auto-instrumentations-node package requires an older version of ioredis but latest apollo-utils requires a newer package of ioredis, leading to typescript compile issues
* we don't even need the opentelemetry package that requires ioredis bc we're not using redis (nor really xray/opentelemetry setups at all)
* as part of getting the above setup, i blew away the lockfile & found that ts-jest && jest updates expected an updated version of typescript, which caused other compilation errors, so downgraded ts-jest (and removed jest so ts-jest just installs the version of jest it needs)

## Implementation Decisions
- replace the autoinstrumentation stuff with specific opentelemetry packages it seems likely we need (though my inclination here really is to simply remove the opentelemetry & xray setup til more of the bugs & goals are worked out)

## References

Jira ticket:
* blocked ticket: https://getpocket.atlassian.net/browse/INFRA-812
* ticket where this particular bug has occurred before with opentelemetry/xray setup (though its framed not as a bug and as a task): https://getpocket.atlassian.net/browse/INFRA-849


sri-kirsh (approved):
thanks for taking this up christina :)  ya we did a similar thing in parser can probably make these changes as we get to client-api, and user-api someday 🤞

src/server/tracing.ts
```diff
@@ -55,22 +61,27 @@ export async function nodeSDKBuilder() {
...
       }),
-      new HttpInstrumentation({
-        ignoreIncomingPaths: ['/.well-known/apollo/server-health'],
-      }),
+      new AwsLambdaInstrumentation({}),
+      new DataloaderInstrumentation({}),
       new ExpressInstrumentation({
         ignoreLayersType: [ExpressLayerType.MIDDLEWARE],
       }),
+      new FsInstrumentation({}),
```
sri-kirsh, line 72: nit: we also removed fs from parser  as it was producing a lot of noise readings between resolver and sql queries metrics

cmharlow:
> thanks for taking this up christina :) ya we did a similar thing in parser can probably make these changes as we get to client-api, and user-api someday 🤞

I'd personally avoid using this opentelemetry & xray setup entirely until we have a better handle on what we want to accomplish with it & done more production-readiness testing. I've never used it / found what is produced useful, and I've noticed a lot of bugs and general noise coming from these setups.

kschelonka (approved):

src/server/tracing.ts
cmharlow: looking now, will remove if not using fs

sri-kirsh:
>I'd personally avoid using this opentelemetry & xray setup entirely until we have a better handle on what we want to accomplish with it & done more production-readiness testing. 

for context: we had a requirement to add telemetry to list/parser-wrapper/user/client-api to gather data on the SQL calls with resolver chainbefore web team went 100% with their traffic. this was not captured in apollo studio metrics. 

agree on the noise: we experimented adding auto-instrumentation..that's why we are moving away from auto-instrumentation..happy to take up any additional suggestion to improve this/move away from xray to get better insights :)

PR #644, opened 2023-02-16
Circleci repairs

cmharlow:
## Goal

Some CircleCI repairs:
* remove circleci cache as we weren't using it (using npm ci)
* align all circleci node docker images - lambda tasks were sometimes using node:14 images (i'm guessing this is just copypasta as the lambda code subdir .nvmrc indicates it should use node 16)


kschelonka (approved):
😎

PR #652, opened 2023-02-21
chore(tag): remove _createdAt and _updatedAt from Tag entity

sri-kirsh:
## Goal
- remove _createdAt and _updatedAt from Tag entity

## Why?
- these fields are not being used by any clients
- however tag's time_created and time_updated in the database contains null values, so some of our queries had to do `min/max` of tags and causes database performance degradation, especially for users with larger tags (e.g 600k, 1M+ tags)

Ticket: https://getpocket.atlassian.net/browse/INFRA-1058

slack message: https://pocket.slack.com/archives/C03QVL4SU/p1677005467804549


docker-compose.yml
```diff
@@ -14,7 +14,7 @@ services:
       - TZ=US/Central
 
   localstack:
-    image: localstack/localstack:1.2
+    image: localstack/localstack:1.3
```
sri-kirsh, line 17: breaks local test - so updated to next version

more info here: https://pocket.slack.com/archives/C03QVL4SU/p1677016106673369

src/dataService/tagDataService.ts
```diff
@@ -130,8 +125,8 @@ export class TagDataService {
       .pluck('tag');
 
     const tags = await this.getTagsByUserSubQuery()
-      .whereIn('tag', latestTags)
-      .orderBy('_updatedAt', 'desc');
+      .whereIn('tag', latestTags);
+      //.orderBy('_updatedAt', 'desc');
```
sri-kirsh, line 129:
```suggestion
```

src/test/graphql/queries/suggested-tags.integration.ts
```diff
@@ -127,8 +126,8 @@ describe('tags query tests - happy path', () => {
       const tags = res.body.data?._entities[0].savedItemById.suggestedTags;
       expect(tags.length).to.equal(3);
       expect(tags[0].name).to.equal('adventure');
-      expect(tags[1].name).to.equal('thriller');
-      expect(tags[2].name).to.equal('romance');
+      expect(tags[1].name).to.equal('romance');
```
sri-kirsh, line 129: coz we removed `order by updatedAt` for suggestedTags, the order is different but the top 3 values are same

cmharlow (approved):
Thanks for doing this!

PR #654, opened 2023-02-22
feat(saveFavorite): mutation happy path and not-found implementation

sri-kirsh:
## Goal
mutation happy path and not-found implementation

## Implementation details/feedback on:
- pattern followed for savesArchive (details in this PR: https://github.com/Pocket/list-api/pull/647)
- consolidated method with shared logic and shared types


src/dataService/pocketSavesService.ts
```diff
@@ -172,6 +184,58 @@ export class PocketSaveDataService {
...
+   * Writes the update to the database
+   * If the value is not found, throws a NOT_FOUND error
+   * Skips the row whose value is already set.
+   * @param updateValues fields to be updated in the database
+   * @param ids savesId to be updated
+   * @param checkField database field to check if the field is already set
+   * @param value skip row if the value matches
+   * @private
+   */
+  private async writeToDatbabase(
```
kschelonka, line 233: nice

src/dataService/utils.ts
```diff
@@ -32,15 +32,7 @@ export function mysqlDateConvert(mysqlDate: Date | string | null): Date | null {
...
-  const keys = new Set();
-  const uniqueValues = [];
-  input.forEach((element) => {
-    if (!keys.has(element)) {
-      keys.add(element);
-      uniqueValues.push(element);
-    }
-  });
-  return uniqueValues;
+  return Array.from(new Set(input));
```
kschelonka, line 35: technically this performs slower but since we're constrained to 30 elements it's fine

src/dataService/pocketSavesService.ts
```diff
@@ -172,6 +184,58 @@ export class PocketSaveDataService {
...
+    updateValues: ListFavoriteUpdate | ListArchiveUpdate,
+    ids: number[],
+    checkField: 'status' | 'favorite',
+    value: FavoriteStatus | PocketSaveStatus
+  ) {
```
kschelonka, line 238: Add response type

src/models/pocketSave.ts
```diff
@@ -80,22 +80,72 @@ export class PocketSaveModel {
...
+   * @param missing missing savesId
+   * @param updated updated saves
+   * @param path mutation path
+   * @private
+   */
+  private fetchSavesAndError(
+    missing: string[],
+    updated: ListResult[],
+    path: GraphQLResolveInfo['path']
+  ) {
```
kschelonka, line 139: add response type

src/models/pocketSave.ts
```diff
@@ -80,22 +80,72 @@ export class PocketSaveModel {
...
+  }
+
+  /**
+   * construct saves and errors types from the data services
+   * @param missing missing savesId
+   * @param updated updated saves
+   * @param path mutation path
+   * @private
+   */
+  private fetchSavesAndError(
```
kschelonka, line 135: This isn't fetching -- maybe rename to `formatSavesMutationPayload` (whatever the graphql entity name is)

src/models/pocketSave.ts
```diff
@@ -80,22 +80,72 @@ export class PocketSaveModel {
...
+      missing.length > 0
+        ? missing.map((missingId) => this.notFoundPayload('id', missingId))
+        : [];
+    const save = updated.map((row) => PocketSaveModel.transformListRow(row));
     const resolvedErrors = errors.map((error) => ({
       ...error,
       path: this.context.models.baseError.path(path),
     }));
-    return { save, errors: resolvedErrors };
+    return { save, resolvedErrors };
```
kschelonka, line 149: This is for the mutation payload response type

```suggestion
    return { save, errors: resolvedErrors };
```

src/models/pocketSave.ts
```diff
@@ -80,22 +80,72 @@ export class PocketSaveModel {
...
+    ids: string[],
+    timestamp: Date,
+    path: GraphQLResolveInfo['path']
+  ): Promise<SaveWriteMutationPayload> {
+    const uniqueIds = uniqueArray(ids.map((id) => parseInt(id)));
+    const { updated, missing } = await this.saveService.favoriteListRow(
+      uniqueIds,
+      timestamp
+    );
+    const { save, resolvedErrors } = this.fetchSavesAndError(
```
kschelonka, line 119: If you use the SaveWriteMutationPayload response type in your function you won't have to extract this and reformat the return type again

src/models/pocketSave.ts
```diff
@@ -80,22 +80,72 @@ export class PocketSaveModel {
...
+  }
+
+  /**
+   * bulk favorite the saves
+   * no action performed if saves are already favorited
+   * any error in the batch will rollback and fail the entire batch
+   * @param ids itemIds associated to the saves to archive; must all be
+   * valid or the operation will fail.
+   * @param timestamp timestamp for when the bulk operation occurred
+   * @returns an array of updated saves and the missing ids (if any
```
kschelonka, line 105: This was probably copied from a different function -- it returns a promise of SaveWriteMutationPayload

src/dataService/utils.ts
sri-kirsh: ya 30 is small amount 
asking for learning: i am seeing both method process list with O(n) and add them to set, and adding to set would be O(1) so they both should take same time right?

kschelonka: I think the time complexity is the same, but I looked at some benchmarking tests about the multiple copying (array + set creation) which was slower than starting with an empty array and pushing values to the set. but as I said I think being readable is preferable here since it's not huge data, so I agree with this change!

kschelonka (approved):

PR #656, opened 2023-02-24
feat(PendingItem): add itemId to pending item

kschelonka:
## Goal
Add itemId to Item and PendingItem

- Resolve the type with the __typename field we already provide from the Item resolver function
- Add ItemModel with method for resolving an Item on a PocketSave
- Use fed2 directives because itemId is now `@shareable` on Item

~Edit: Need a little work to get the schema to compose after the fed2 move, but all other code is ready~ Fixed

- Including on PocketResource

## Implementation Decisions
- I added the item field to PocketSave, although it will be superseded by `resource` according to the contract. I can remove it if we want.

[INFRA-1040]

[INFRA-1040]: https://getpocket.atlassian.net/browse/INFRA-1040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/test/graphql/mutations/savedItemsMutationService-upsert.integration.ts
+++ b/src/test/graphql/mutations/savedItemsMutationService-upsert.integration.ts
@@ -210,9 +210,12 @@ describe('UpsertSavedItem Mutation', () => {
             id
             item {
               ... on Item {
+                __typename
                 givenUrl
               }
               ... on PendingItem {
+                __typename
+                itemId
                 url
               }
             }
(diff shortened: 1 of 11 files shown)


bassrock (approved):
TY!

src/test/graphql/mutations/savedItemsMutationService-upsert.integration.ts
```diff
@@ -229,6 +232,8 @@ describe('UpsertSavedItem Mutation', () => {
       expect(data.id).to.equal('1');
       expect(data.item.givenUrl).is.undefined;
       expect(data.item.url).to.equal(givenUrl);
+      expect(data.item.itemId).to.equal('1');
```
kschelonka, line 235: @bassrock here's where we're testing the resolution for savedItem

PR #663, opened 2023-03-01
feat(unfave): ability to unfavorite a pocketsave

cmharlow:
## Goal

Adds a mutation to unfavorite a PocketSave entity.

Does not add emission of events yet given separate discussion that moved that to another ticket.

i was half tempted to simply add a 'favorite/unfavorite' argument to the existing favoriteListRow && saveFavorite methods, but decided to go more explicit (and verbose). but not opposed to merging these however.

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-546


kschelonka (approved):
eh, a little verbosity is fine. lgtm 🦅

PR #664, opened 2023-03-01
feat(tags): saveBatchUpdateTags mutation [INFRA-809]

kschelonka:
## Goal

saveBatchUpdateTags mutation

- Implement saveBatchUpdateTags mutation
- Create a specific NotFound class which implements BaseError -- enables reusing the message payload across the Tag and PocketSave models
- rename TagCreateInput to SaveTagAssociation (so it can be reused without having a confusing namespace, as it's actually pretty generic)
- add optional timestamp inputs to some saveditem data service methods, so that they can be reused without refactoring any other code

Apologies for the large amount of files -- I made some small tweaks to more easily reuse things like an input type, some data services functions, a more specific class for NotFound, etc. Have commented to help direct the review.

- Potential for large transactions. Even with a 30-item constraint on each of the array fields,the total number of possible writes to the database in a single request is (30 [creates] + 30 [deletes]) * 30 [inputs] = 1800. That seems like a lot... Maybe we could consider a total node limit in addition to (or in place of) some of the proposed constraints?
- Could consider breaking the transaction up per-save to limit big transactions, but then the app would have to be in charge of ensuring proper rollback...
- Tag limits? Right now we don't enforce limits on how many tags can be applied to an item. There's probably a reasonable number here. Would need product buy-in and would be outside this scope.

## Implementation Decisions
- I put the methods on the tag model because of the tag input validation, but like all tag stuff it returns a save (and the tags are technically more like an association entity)
- deleting a tag that doesn't exist is a no-op, not an error
- adding a tag that already exists is a no-op
- use the pre-existing methods for database writes as much as possible

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-809


src/dataService/savedItemsService.ts
```diff
@@ -379,8 +379,11 @@ export class SavedItemDataService {
    * the list table, by item id.
    * @param itemIds The item IDS to update the `time_Updated` to now.
    */
-  public updateListItemMany(itemIds: string[]): Knex.QueryBuilder {
-    return this.listItemUpdateBuilder().whereIn('item_id', itemIds);
+  public updateListItemMany(
```
kschelonka, line 382: Add an optional timestamp to these functions so they can be reused (without having to refactor any other code)

src/dataService/tagDataService.ts
```diff
@@ -182,31 +186,42 @@ export class TagDataService {
...
     await this.db.transaction(async (trx: Knex.Transaction) => {
       await this.insertTagAndUpdateSavedItem(tagInputs, trx);
       await this.usersMetaService.logTagMutation(new Date(), trx);
     });
   }
 
   private async insertTagAndUpdateSavedItem(
-    tagInputs: TagCreateInput[],
-    trx: Knex.Transaction<any, any[]>
+    tagInputs: TagSaveAssociation[],
```
kschelonka, line 197: small tweaks to add an optional timestamp field, linter reformatting

src/models/notFoundError.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseErrorModel } from './baseError';
+import { NotFoundInternal } from '../types';
+
+export class NotFoundErrorModel extends BaseErrorModel {
+  public message(key: string, value: string): NotFoundInternal {
```
kschelonka, line 5: Found msyelf copy-pasting code, so I made a more specific implementation of the BaseError class

src/models/tag.ts
```diff
@@ -230,11 +235,12 @@ export class TagModel {
...
-      replacement.tags.map((tag) => ({
-        savedItemId: replacement.savedItemId,
-        name: sanitizeTagName(tag),
-      }))
+    const tagCreates: TagSaveAssociation[] = replacements.flatMap(
+      (replacement) =>
+        replacement.tags.map((tag) => ({
+          savedItemId: replacement.savedItemId,
+          name: sanitizeTagName(tag),
+        }))
```
kschelonka, line 243: everything above this line is just rename

src/models/tag.ts
```diff
@@ -243,6 +249,92 @@ export class TagModel {
...
+  ): Promise<SaveWriteMutationPayload> {
+    const batchUpdate = this.formatBatchUpdateInput(input);
+    const { updated, missing } = await this.tagService.batchUpdateTags(
+      batchUpdate,
+      timestamp
+    );
+    let save: PocketSave[] = [];
+    if (updated.length) {
+      const ids = updated.map((update) => update.toString());
+      save = await this.context.models.pocketSave.getManyById(ids);
```
kschelonka, line 277: could use DI instead

src/resolvers/utils.ts
```diff
@@ -1,5 +1,9 @@
 import DataLoader from 'dataloader';
-import { SavedItemTagsInput, SavedItemTagsMap, TagCreateInput } from '../types';
+import {
```
kschelonka, line 2: all renames and linting

src/dataService/tagDataService.ts
```diff
@@ -182,31 +186,42 @@ export class TagDataService {
...
+    trx: Knex.Transaction<any, any[]>,
+    timestamp?: Date
   ) {
-    const timestamp = mysqlTimeString(new Date(), config.database.tz);
+    if (tagInputs.length === 0) {
```
cmharlow, line 201: wouldn't this be caught by constraints? but i guess they're commented out atm

src/dataService/tagDataService.ts
```diff
@@ -182,31 +186,42 @@ export class TagDataService {
...
         api_id: parseInt(this.apiId),
       };
     });
     await trx('item_tags').insert(inputData).onConflict().ignore();
-    const itemIds = tagInputs.map((element) => element.savedItemId);
-    await this.savedItemService.updateListItemMany(itemIds).transacting(trx);
+    const itemIds = uniqueArray(
+      tagInputs.map((element) => element.savedItemId)
+    );
+    await this.savedItemService
```
cmharlow, line 222: so this essentially just updates the updatedAt time for the savedItem because of the previously success tag association creation? Am I reading that correctly?

src/types/index.ts
```diff
@@ -239,3 +239,21 @@ export type SaveMutationInput = {
...
+  removeTagIds: string[];
+  addTagNames: string[];
+};
+
+/**
+ * Input in a better format for doing bulk transactions. Group
+ * into array of creates and deletes.
+ * Transform the tagIds into "names", which corresponds to the
+ * item_tags.tag field in the database and is part of the primary key
+ * (which also includes user_id and item_id).
```
cmharlow, line 254: user_id included only bc of the user_id + item_id == savedItemId approach, right? (just trying to map the definition to what I see defined here)

src/dataService/tagDataService.ts
```diff
@@ -373,4 +426,16 @@ export class TagDataService {
       .where({ user_id: this.userId, tag: tagName })
       .del();
   }
+
+  /** Query builder for deleting item_tags rows by PK tuples */
+  private deleteTagsByNameAndItemId(
+    association: TagSaveAssociation[]
+  ): Knex.QueryBuilder {
+    const tuples = association.map(({ savedItemId, name }) => {
+      return [parseInt(savedItemId), name, this.userId];
```
cmharlow, line 435: ah i ask a question about user_id later, sorry i was reading out of order

src/dataService/tagDataService.ts
```diff
@@ -362,6 +377,44 @@ export class TagDataService {
...
+      [...deletes, ...creates].map((req) => parseInt(req.savedItemId))
+    );
+    const missing = await this.pocketSaveService.checkIdExists(saveIds);
+    // Exit if there are any missing Saves, with the missing ids returned
+    if (missing.length) {
+      return { updated: [], missing: missing.map((id) => id.toString()) };
+    }
+    // Otherwise we can proceed with the write
+    // We won't throw error if trying to delete a nonexistent tag (no-op)
+    await this.db.transaction(async (trx) => {
```
cmharlow, line 410: this is all a single transction, if i'm reading this correctly (and recalling correctly the PR descriptino, which I'm going to double check after I get through this first review - apologies as you've already explained this I believe)?

It does seem like it could be a rather large single transaction if so. I wonder if we could do some load testing of this, to see what upper limits of db hardware resourcing we could possibly hit. Not enough to warrant pausing this PR, just thinking out loud.

src/models/baseError.ts
```diff
@@ -1,4 +1,5 @@
 import { GraphQLResolveInfo } from 'graphql';
+import { NotFoundInternal } from '../types';
```
cmharlow, line 2:
```suggestion
```

Just added in light of linter error / linter report this import isn't used.

src/models/pocketSave.ts
```diff
@@ -58,6 +49,23 @@ export class PocketSaveModel {
     return pocketSave;
   }
 
+  /**
```
cmharlow, line 52: this covers this ticket? https://getpocket.atlassian.net/browse/INFRA-810

src/models/tag.ts
```diff
@@ -2,20 +2,25 @@ import {
...
+  SaveUpdateTagsInputGraphql,
+  SaveUpdateTagsInputDb,
+  SaveWriteMutationPayload,
 } from '../types';
 import config from '../config';
 import { IContext } from '../server/context';
 import { SavedItemDataService, TagDataService } from '../dataService';
 import { NotFoundError, UserInputError } from '@pocket-tools/apollo-utils';
 import { addslashes } from 'locutus/php/strings';
+import { uniqueArray } from '../dataService/utils';
```
cmharlow, line 21:
```suggestion
```
just proposal from linter feedback that this isn't used.

src/models/tag.ts
```diff
@@ -243,6 +249,93 @@ export class TagModel {
...
+  ): Promise<SaveWriteMutationPayload> {
+    const batchUpdate = this.formatBatchUpdateInput(input);
+    const { updated, missing } = await this.tagService.batchUpdateTags(
+      batchUpdate,
+      timestamp
+    );
+    let save: PocketSave[] = [];
+    if (updated.length) {
+      const ids = updated.map((update) => update.toString());
+      save = await this.context.models.pocketSave.getManyById(ids);
```
cmharlow, line 277: ah why the above that I thought was https://getpocket.atlassian.net/browse/INFRA-810 was setup - i get it

src/models/tag.ts
cmharlow: DI?

cmharlow: oh, dependency injection?

cmharlow (approved):
alright, I read through everything the best I could, and I'm running it locally now to try to continue to build my understanding of the changes here. I left some comments, mostly questions to aide my understanding.

I won't block anything bc it makes sense to me so far. My main concern is large transactions given recent db resource concerns around large, unclosed transactions. 

For that to happen though, it would necessitate - i think - something trying to do a large out of both creates & deletes of tag associations for a lot of items... I think? Or an inordinately large amount of one of those vectors. 

With regards to rate limiting, I'm clearly not the best person (I'm probably the worst person on the team atm) to give insights, but I will say being  a heavy user of GitHub's cli & GraphQL endpoint, I find their approach understandable from a client pov: https://docs.github.com/en/graphql/overview/resource-limitations This is scoring based, indirectly, on essentially number of nodes you might hit with a particular query - so in your PR description, limiting number of nodes. I like that approach, perhaps in tandem with constraints until we're more comfortable with our setup, monitoring & usage of this service, because it protects us from those singular but inordinately large vectors - e.g. a user with maybe a few items but each item has hundreds of tags. We could also do some analysis (perhaps yall already have) to think through the risk areas like this though.

schema.graphql
```diff
@@ -109,6 +109,20 @@ input PaginationInput {
   last: Int
 }
 
+input SaveUpdateTagsInput {
+  saveId: ID!
+  """
+  Tags to remove, by ID
+  """
+  removeTagIds: [ID!]! #@constraint(maxItems: 30)
```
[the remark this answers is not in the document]
kschelonka: yep, just have them as placeholders until we can do https://getpocket.atlassian.net/browse/INFRA-1036

src/dataService/tagDataService.ts
kschelonka: yeah, more a caution since this is a private function that isn't specifically tied to that resolver

src/dataService/tagDataService.ts
kschelonka: yes, happy to explain more later... it's how offline-capable (mobile) clients sync right now. If the save timestamp is updated they refresh all the other data associated with it (e.g. tags)

The lack of a more robust sync system is also the reason we can't hard-delete anything right now.

src/models/pocketSave.ts
kschelonka: Yes, there is some overlap here 😳 . I imagine the resolver would call this method, but the schema still needs to be updated and integration tests for that specific resolver added.

kschelonka: Oh and the schema has to be checked -- as per my comment I think that resolver might return a union type with NotFound.

src/dataService/tagDataService.ts
kschelonka: Yeah absolutely, we both have concerns about this. I'm wondering if a node limit would be good to enforce? But yes I'm not sure what that limit should be, which we probably need to do empirically...

src/types/index.ts
kschelonka: Thank you, this comment is worded poorly. What I was trying to say is that the item_tags primary key is `(tag, user_id, item_id)`. So transform the input to extract the tag name from the "id" input (which at this time is constructed from the tag name `item_tags.tag`, since there isn't a concept of a tag id.)

src/dataService/tagDataService.ts
kschelonka: I went ahead and added a 150-node write limit (delete/create) as a starting point for testing.

cmharlow (approved):

PR #666, opened 2023-03-02
feat(unarchive): ability to unarchive a pocketsave

cmharlow:
## Goal

Add ability to unarchive a PocketSave or list of PocketSaves.

Also set the app TZ for the docker-compose setup to US/Central as well, as it was showing errors running locally (undesired timezone shifts). See impl notes for more details on this.

- again, the main push of this PR is just copy & pasting & altering with a 'un' the archive stuff already existing, but i'm open to cutting down on verbosity & just adding on/off arguments to existing functions

## Implementation Decisions
- i did investigate whether or not we could get away from forcing all app servers using this code to be in the same TZ (system-wide) by using the MySQL package ConnectionConfig.timezone option. Unfortunately, it just doesn't work with negative offsets. Plus with this route, we'd have to manage DST changes ourselves. It  would obviate the need for setting system TZs everyone, but unfortunately, isn't currently usable.

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-544

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,6 +53,7 @@ services:
       - .env
     environment:
       - NODE_ENV=local
+      - TZ=US/Central
     entrypoint: ./.docker/scripts/local-start.sh
     depends_on:
       - mysql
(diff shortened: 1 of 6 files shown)


jpetto (approved):
the copy pasta isn't too egregious imo.

Herraj (approved):

PR #684, opened 2023-03-16
feat(tags): use sort key as part of cursor [INFRA-1057]

kschelonka:
## Goal

Use the new unique sort key as part of the cursor for paginating tags. this helps because most of the time due to other legacy bugs, the timestamp columns are null (so there's nothing to sort by). 

HOWEVER - there are still a number of issues around tags, including:
* returning a potentially huge amount of aggregate array data (item_ids associated to a tag, due to no limits on list size)
* potential for massive transactions due to lack of proper data models and sync design (https://pocket.slack.com/archives/C03QVL4SU/p1678979927687699)
* still potential for poor read performance since all tags queries require aggregation, and we have no limits on list size

**Changes**
- Use the new unique sort key as part of the cursor for paginating tags
- Test data overhaul; with the new sort key all the tags tests had to be updated

- nothing specifically here, but can take to slack for other issues laid out with tags

## Implementation Decisions
- Didn't remove the concatenated list response of all saves associated with a tag
    - these values were required by other mutations (e.g. renaming a tag), and I didn't want to blow up the scope
-  Slowly removing chai in favor of jest 

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-1057


src/dataService/tagDataService.ts
```diff
@@ -51,11 +51,12 @@ export class TagDataService {
       .select(
         'tag as name',
         'tag',
+        this.db.raw('MAX(id) as _cursor'),
```
sri-kirsh, line 54: if we need a sortId to be mapped with the tag's cursor, would it make sense to get the first tag's cursor, so we don't parse the entire tag list of the user?

_(i might be missing some context on why we need to do max)_ may be, we can document why we chose max as comment

src/test/graphql/queries/tags.integration.ts
```diff
@@ -387,51 +366,58 @@ describe('tags query tests - happy path', () => {
...
+      tags: [
+        expect.objectContaining({
+          name: 'travel',
+          id: toBeStringOfLengthGreaterThanOne(),
+          savedItems: expect.objectContaining({
+            edges:
+              // 4 saves with this tag, and one should be the parent save url
+              expect.toBeArrayOfSize(4) &&
+              expect.arrayContaining([
+                expect.objectContaining({
```
sri-kirsh, line 387: 👍

sri-kirsh (approved):

src/dataService/tagDataService.ts
kschelonka: using max here gets the tags in order they were most recently used. so if I have tagging behavior like this:

(ascending order from past) 

tag item A as 'recipe'
tag item B as 'work'
tag item C as 'recipe'

A.id < B.id < C.id

but in aggregate, 'recipe'.id > 'work'.id, so tags in most recently used order would show up appropriately


what you're describing is a limitation in the modeling... I agree here that it would be best if we didn't have to group all the same-named tags

PR #711, opened 2023-04-03
feat(upsertSavedItem): [INFRA-1195] Add title field to upsertSavedItem mutation

Herraj:
## Goal
Adding support for the `title` field for the `upsertSavedItem` mutation.

We can now provide `title` as part of the mutation input. If provided in the input it will be saved and returned. 
If not provided as part of the mutation input, will be set to what the parser call returns and saved as that.

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-1195

### DIFF
--- a/.circleci/scripts/setup_aws.sh
+++ b/.circleci/scripts/setup_aws.sh
@@ -5,7 +5,7 @@ sudo apt-get update && sudo apt-get install -y python3-pip
 # 2022-12-01 - awscli is pinned to fix a build error with 1.27.1 related to
 # not finding a botocore version = 1.29.21. this can probably be
 # un-pinned in the future?
-pip3 install awscli-local awscli==1.27.20
+pip3 install boto3==1.26.90 awscli-local awscli==1.27.20
 
 
 for Script in .docker/localstack/*.sh ; do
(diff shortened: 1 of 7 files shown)


.circleci/scripts/setup_aws.sh
```diff
@@ -5,7 +5,7 @@ sudo apt-get update && sudo apt-get install -y python3-pip
 # 2022-12-01 - awscli is pinned to fix a build error with 1.27.1 related to
 # not finding a botocore version = 1.29.21. this can probably be
 # un-pinned in the future?
-pip3 install awscli-local awscli==1.27.20
+pip3 install boto3==1.26.90 awscli-local awscli==1.27.20
```
Herraj, line 8: Ticket created for this: https://getpocket.atlassian.net/browse/INFRA-1202

src/resolvers/mutation.ts
```diff
@@ -35,10 +35,16 @@ export async function upsertSavedItem(
   const savedItemDataService = new SavedItemDataService(context);
 
   try {
-    const item = await ParserCaller.getOrCreateItem(savedItemUpsertInput.url);
+    //TODO do we need the resolved id @Herraj
```
sri-kirsh, line 38: @kschelonka this came up during pairing. even though we are not exposing resolved_id through pocket graph's `savedItem`, we still populate resolved_id to maintain backward compatibility with web repo, right? 

otherwise, we were thinking to skip the parser call if the title is provided to prevent a downstream call

sri-kirsh (approved):

PR #717, opened 2023-04-07
feat(constraint): add constraint directive

kschelonka:
## Goal
Add constraint directives to list-api schema.

- Update federation version and imports
- Add constraint directive and compose
- Update server to include constraint defs in schema
- Update build process to generate schema in /dist (build folder)
- Update circleci to use generated schema

- any

## Implementation Decisions
- generate schema in /dist (build folder)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,20 +64,20 @@ jobs:
           name: install rover
           command: |
             # download and install Rover
-            curl -sSL https://rover.apollo.dev/nix/v0.1.0 | sh
+            curl -sSL https://rover.apollo.dev/nix/v0.11.1 | sh
 
             # This allows the PATH changes to persist to the next `run` step
             echo 'export PATH=$HOME/.rover/bin:$PATH' >> $BASH_ENV
       - run:
           name: check service
-          command: rover subgraph check pocket-client-api@current --schema ./schema.graphql --name=list
+          command: rover subgraph check pocket-client-api@current --schema ./dist/schema-generated.graphql --name=list
       - when:
           condition:
             equal: [main, << pipeline.git.branch >>]
           steps:
             - run:
                 name: push service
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -57,27 +57,29 @@ only_dev: &only_dev
 jobs:
   apollo:
     docker:
-      - image: circleci/node:16.13.1@sha256:1803e9ed7deec9456ad2609124b7333d40b2eec0cf34998ae766cbf90c9a3625
+      - image: cimg/base:2023.04
```
kschelonka, line 60: a more minimal image since we don't need node for this step

cmharlow (dismissed):
few questions about the CI and some perhaps unintentional OS changes / environment hard-coding in setup.

.circleci/config.yml
kschelonka: Rover just invokes the managed federation steps we have in apollo studio -- I don't think the OS should matter as long as the tool works

.circleci/config.yml
```diff
@@ -57,27 +57,29 @@ only_dev: &only_dev
...
             # download and install Rover
-            curl -sSL https://rover.apollo.dev/nix/v0.1.0 | sh
+            curl -sSL https://rover.apollo.dev/nix/v0.11.1 | sh
 
             # This allows the PATH changes to persist to the next `run` step
             echo 'export PATH=$HOME/.rover/bin:$PATH' >> $BASH_ENV
       - run:
           name: check service
-          command: rover subgraph check pocket-client-api@current --schema ./schema.graphql --name=list
+          command: rover subgraph check pocket-client-api@current --schema /tmp/workspace/app_prod/dist/schema-generated.graphql --name=list
```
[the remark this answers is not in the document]
kschelonka: we don't have a conditionally set workspace, so it won't make a difference. this is just the circleci directory shared between jobs

cmharlow (dismissed):

cmharlow (approved):

PR #720, opened 2023-04-10
bug(dataloader): use dataloaders for tags by ItemId to avoid n+1 problem

cmharlow:
## Goal

Sets up some problematic, db-querying functions - see description in [ticket here](https://getpocket.atlassian.net/browse/INFRA-1056) - to use a dataloader (without cache enabled, as mutations rely on this codepath at times) to leverage the simplest form of batching. This batching groups the multiple tags per ItemId database queries into 1 tags per ItemIds query (using an `itemId in` versus `itemId = ` sql query for batching). This then cuts down on the used database connections, helping with the original error that's been plaguing list-api - out of connections when trying to get all tags for all items for a given user.

This only uses a Tags-specific dataloader for tagsByItemIds for now - while there are other Tags dataloaders (for tags by id, tags by name), those aren't used yet. This doesn't add a dataloader for Tags by User Id yet, but that could be a next step (we'd need to mix the existing pagination pattern with the desired dataloader pattern - probably not too hard, but more cognitive load than desired for this first PR).

The dataloaders are left in / added to the dataloaders subdir. The tags by Item Ids dataloader works for SavedItem & PocketSave entities. We'll probably want to refactor where what code lives to start to align the models - dataservice patterns set up for PocketSave to encompass dataloaders, but this PR doesn't do that. 

Adds / updates some tests to better capture the dataloader is being used & the batch functions added work as expected.

Makes the database query return distinct Tags (as multiple Item Ids could return the same Tags, and we don't expect duplicate Tags beyond the Data Service layer yet, so handling deduping there via SQL via knex).

For some reason, SavedItem mutations that require a tags lookup will pass the Item ID as an int instead of a string, so also ensure the batch tags function forced ItemIDs to string in the filtering step.

Adds a basic dataloader max batch size function to avoid sending the database queries that return massive responses. Expect this will be tweaked as we see this in action. Also adds a comment for doing our own scheduling of the batch runs - [instead of the default action that will occur with this PR as is](https://github.com/graphql/dataloader/#batch-scheduling) - for future discussion. Played with this a bit, but testing timeouts make me want to revisit this when we have better sense of desired and expected latencies here.

## References

Jira ticket:
* https://getpocket.atlassian.net/browse/INFRA-1056

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -16,31 +16,6 @@ export default {
     environment: process.env.NODE_ENV || 'development',
     depthLimit: 8,
   },
-  events: {
-    source: serviceName, // TODO - ok to change from 'backend-php'?
-    version: '0.0.2', // TODO - version currently in documentation
-  },
-  data: {
-    // A suffix that ensures the tag ID is never an empty string,
-    // because there are empty tag strings in the DB
-    tagIdSuffix: '__xpktxtagx__',
-  },
-  awsEnvironments,
-  snowplow: {
-    endpoint: process.env.SNOWPLOW_ENDPOINT || 'localhost:9090',
-    httpProtocol: snowplowHttpProtocol,
-    bufferSize: 1,
-    retries: 3,
-    appId: 'pocket-backend-list-api',
-    events: EventType,
-    schemas: {
-      listItemUpdate: 'iglu:com.pocket/list_item_update/jsonschema/1-0-1',
-      listItem: 'iglu:com.pocket/list_item/jsonschema/1-0-1',
-      content: 'iglu:com.pocket/content/jsonschema/1-0-0',
-      user: 'iglu:com.pocket/user/jsonschema/1-0-0',
-      apiUser: 'iglu:com.pocket/api_user/jsonschema/1-0-0',
-    },
-  },
   aws: {
     region: process.env.AWS_REGION || 'us-east-1',
     endpoint: localAwsEndpoint,
@@ -86,6 +61,25 @@ export default {
       batchSize: 10,
     },
   },
+  awsEnvironments,
+  batchDelete: {
+    deleteDelayInMilliSec: 500,
+    tablesWithPii: ['item_tags', 'list', 'item_attribution'],
+    tablesWithUserIdAlone: [
+      'list_meta',
+      'items_scroll',
+      'item_ads',
+      'item_time_spent',
+      'item_currently_reading',
+      'list_extras',
+      'list_shares',
+    ],
+  },
+  data: {
+    // A suffix that ensures the tag ID is never an empty string,
+    // because there are empty tag strings in the DB
+    tagIdSuffix: '__xpktxtagx__',
+  },
   database: {
     // contains tables for user, list, tags, annotations, etc.
     read: {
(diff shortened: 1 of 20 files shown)


src/config/index.ts
```diff
@@ -103,37 +97,47 @@ export default {
     dbName: process.env.DATABASE || 'readitla_ril-tmp',
     tz: process.env.DATABASE_TZ || 'US/Central',
   },
-  sentry: {
-    dsn: process.env.SENTRY_DSN || '',
-    release: process.env.GIT_SHA || '',
-    environment: process.env.NODE_ENV || 'development',
+  dataloaderDefaults: {
```
cmharlow, line 100: this is the only real addition to this file in this PR. The rest is putting the config values in alphabetical order (at the top level) to understand better what exists vs doesn't.

src/dataLoader/tagsDataLoader.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+    sinon.stub(service, 'batchGetTagsByUserItems').returns(promiseTags);
+    const tags = await batchGetTagsByItemIds(service, ['1', '3']);
+
+    // dataloader wants an array that is the same length
+    // of array of args passed in
+    // (this is how it matches up results to batched inputs)
+    expect(tags).toHaveLength(2);
+    // there are 2 items with the first given tag;
+    expect(tags[0]).toHaveLength(2);
+    expect(tags[0]).toContainEqual({
```
kschelonka, line 76: don't have to change this, but for future reference this is why I installed jest-extended. you can do `expect.toHaveSameMembers` to assert that an array contains every expected element (and none others).

src/dataLoader/tagsDataLoader.ts
```diff
@@ -1,55 +1,110 @@
...
+    const tags = await batchGetTagsByIds(
+      new TagDataService(context, savedItemdService),
+      ids
+    );
     tags.forEach((tag) => byNameLoader.prime(tag.name, tag));
     return tags;
   });
   const byNameLoader = new DataLoader(async (names: string[]) => {
-    const tags = await batchGetTagsByNames(context, names);
+    const savedItemdService = new SavedItemDataService(context);
```
kschelonka, line 37:
```suggestion
    const savedItemDataService = new SavedItemDataService(context);
```

src/dataLoader/tagsDataLoader.ts
```diff
@@ -1,55 +1,110 @@
...
+    async (itemIds: string[]) => {
+      const savedItemdService = new SavedItemDataService(context);
+      const tags = await batchGetTagsByItemIds(
+        new TagDataService(context, savedItemdService),
+        itemIds
+      );
+      return tags;
+    },
+    {
+      maxBatchSize: config.dataloaderDefaults.maxBatchSize,
```
kschelonka, line 55: do you need `cache=False` here for it to not cache?

kschelonka (approved):

src/dataLoader/tagsDataLoader.spec.ts
cmharlow: yeah, didn't seem like a huge lift for this for now. but noted, thanks

src/dataLoader/tagsDataLoader.ts
cmharlow: the cache push / check is in a diff util function that this batch function doesn't use iirc; will confirm.

cmharlow: yeah okay - so for the other tags dataloaders, they call a `tags.forEach((tag) => byIdLoader.prime(tag.id, tag));` in their dataloader function, which then writes to the cache setup by default by dataLoader class. 

I don't do any of that for the byItemId dataloader - so it does look like a cache is created for it by DataLoader class by default, but we just never write to it. Going ahead and also explicitly disabling even making the unused cache to start with too.

kschelonka (approved):

PR #733, opened 2023-04-20
feat(ci): circleci slack notification on failed job steps

kschelonka:
## Goal

Add job step failure notifications. Uses post-steps in circleci to notify on failure of any step. Since the branch filtering is applied to main only, I didn't apply to any steps not run on main.

Tested on previous (reverted) commit using build pass notification to ensure that the integration was properly set up. 

- any

## Implementation Decisions
- use post-steps to hook into each workflow job

Documentation:
* [Project doc](https://getpocket.atlassian.net/wiki/spaces/PE/pages/2950857054/Circleci+Slack+Deployment+Notifications)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,9 @@ jobs:
           root: /tmp/workspace
           paths:
             - app_prod
+      - slack/notify:
+          event: pass
+          template: success_tagged_deploy_1
 
   test_integrations:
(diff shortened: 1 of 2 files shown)


cmharlow (approved):
lgtm, with note (made in person) for adding this to the backstage template bc its annoying to include in all specific jobs

PR #753, opened 2023-05-12
feat(tags): enable tagging by givenUrl

kschelonka:
## Goal
Resolve issues with offline mobile clients by enabling tagging by givenUrl. Other mutations/queries will follow. This is because givenUrl is available offline, but a server-provided itemId (which is the key for accessing all saves at this time), is not available for offline clients creating new saves.

- [x] create new mutation for tagging by givenUrl `savedItemTag`
- [x] emit add_tag events on success
- [x] update parser caller to use the getItem endpoint and extract the itemId from response; small refactor for reusing code and test mocks

- is the parser dependency ok for a short term solution?
- failing with NOT_FOUND if parser does not have an itemId for url (vs. using createIfNone param)

## Implementation Decisions
- Use the parser for resolving itemId instead of replicating the url transformation/hashing logic and looking up in `readitla_b` from list
- alternatively also, use the parser instead of looking up by givenUrl in `list` since givenUrl is not indexed (request has been made to index this field)
- do not create the itemId if it doesn't exist; return NOT_FOUND instead. The itemId should always exist prior to the tag action, and I think there's a risk of having records not connected properly if we use the `createIfNone` method on the parser (potentially the url could have one itemId for the tag records and a different one for the save, depending on some edge conditions). 

## References

Jira ticket:
* [IN-1453]

[IN-1453]: https://getpocket.atlassian.net/browse/IN-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dataService/tagDataService.ts
+++ b/src/dataService/tagDataService.ts
@@ -208,10 +208,14 @@ export class TagDataService {
    * is actually in the user's list (no foreign key constraint).
    * @param tagInputs
    */
-  public async insertTags(tagInputs: TagSaveAssociation[]): Promise<void> {
+  public async insertTags(
+    tagInputs: TagSaveAssociation[],
+    timestamp?: Date
+  ): Promise<void> {
+    const updatedTime = timestamp ?? new Date();
     await this.db.transaction(async (trx: Knex.Transaction) => {
-      await this.insertTagAndUpdateSavedItem(tagInputs, trx);
-      await this.usersMetaService.logTagMutation(new Date(), trx);
+      await this.insertTagAndUpdateSavedItem(tagInputs, trx, updatedTime);
+      await this.usersMetaService.logTagMutation(updatedTime, trx);
     });
   }
 
--- a/src/externalCaller/parserCaller.ts
+++ b/src/externalCaller/parserCaller.ts
@@ -72,7 +72,8 @@ export class ParserCaller {
 
   /**
    * Get the Parser-generated itemId for a given url.
-   * TODO: stop using this method once given_url is indexed in the list table
+   * TODO[IN-1478]: stop using this method once given_url is indexed in the list table
+   * https://getpocket.atlassian.net/browse/IN-1478
    * @param url the URL of the Save you want an itemId for
    * @param tries # of request retries if there are issues with the Parser service
    * @returns the itemId for the given url, or null if it does not exist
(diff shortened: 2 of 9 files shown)


src/externalCaller/parserCaller.ts
```diff
@@ -1,5 +1,13 @@
 import fetch from 'node-fetch';
 import config from '../config';
+import { setTimeout } from 'timers/promises';
+
+export type PartialGetItemResponse = {
```
efixler, line 5: This makes sense but looking at this against the type below makes me want to double check the casing of `item_id` (which is `itemId` below).

src/dataService/tagDataService.ts
```diff
@@ -208,10 +208,14 @@ export class TagDataService {
...
-  public async insertTags(tagInputs: TagSaveAssociation[]): Promise<void> {
+  public async insertTags(
+    tagInputs: TagSaveAssociation[],
+    timestamp?: Date
+  ): Promise<void> {
+    const updatedTime = timestamp ?? new Date();
     await this.db.transaction(async (trx: Knex.Transaction) => {
-      await this.insertTagAndUpdateSavedItem(tagInputs, trx);
-      await this.usersMetaService.logTagMutation(new Date(), trx);
+      await this.insertTagAndUpdateSavedItem(tagInputs, trx, updatedTime);
```
efixler, line 217: Is it correct to assume that ^^ is a backstop to amelioriate the missing dates we see in the db?

efixler:
@kat lgtm. 🎩 for the integration tests kind of a lot of effort for those tests but it's important.

One idiot-check question. 

Your choices sound sensible. Regarding the pending index (second bullet) might be good to leave a comment in the code with a ticket for that. I understand that the index is weeks out atm.

efixler (approved):
Couple questions inline.

src/dataService/tagDataService.ts
kschelonka: no, this is so that we allow clients to set the time for better support for offline devices. otherwise it just sets the time the server received the request.

src/externalCaller/parserCaller.ts
kschelonka: 👍 yep the response from the parser is item_id, but snake_casing isn't idiomatic JS/TS so that's why it's always itemId in const vars/etc.

kschelonka:
> Your choices sound sensible. Regarding the pending index (second bullet) might be good to leave a comment in the code with a ticket for that. I understand that the index is weeks out atm.

good idea, I will add a ticket to track and update the TODOs

kschelonka:
added the ticket number and additional todos, @efixler mind taking another look?

src/externalCaller/parserCaller.ts
efixler: I understand -- here we have 2 types one called `PartialGetItemResponse` the other called `ItemResponse` just below it where one sounds like a Partial of the other (even if it isn't) and they each have the same property in snake/camel variants.

PR #765, opened 2023-05-22
fix(db): chunk queries to avoid full table scan

kschelonka:
If 'where in' array is too large, then
mysql will ignore indices and scan the
entire table. The exact value at which
this happens depends on some memory settings.
Manual testing showed that 10k size will work
(we can tune this with the config value in
the future).


cmharlow (approved):
preemptive okay waiting for additional testing

cmharlow (approved):
Thank you!

PR #801, opened 2023-06-01
feat(archive/unarchive): add archive/unarchive mutation by URL

kschelonka:
## Goal
Add archive/unarchive mutations for SavedItems by Url, to support offline mobile clients

## I'd love feedback/perspectives on:
- any

## Implementation Decisions
- Combining archive/unarchive together since they're so similar
- Moving shared code to a new SavedItem model
- Continuing pattern of using Parser to resolve the SavedItem url, until the givenUrl field can be indexed in `list` table, to avoid potentially huge lookups when users have millions of items

## References

Jira ticket:
* [IN-1472]
* [IN-1476]

[IN-1472]: https://getpocket.atlassian.net/browse/IN-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IN-1476]: https://getpocket.atlassian.net/browse/IN-1476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #802, opened 2023-06-01
feat(favorite/unfavorite): un/favorite by givenUrl

kschelonka:
## Goal
Add mutation to favorite/unfavorite by url identifier, to support offline mobile clients.
Extends the same pattern as #801

## Implementation Decisions
- Continuing the same pattern as archive/unarchive in #801 

## References

Jira ticket:
* [IN-1473]
* [IN-1471]

[IN-1473]: https://getpocket.atlassian.net/browse/IN-1473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[IN-1471]: https://getpocket.atlassian.net/browse/IN-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dataService/savedItemsService.ts
+++ b/src/dataService/savedItemsService.ts
@@ -177,13 +177,16 @@ export class SavedItemDataService {
    * in the table ('time_updated', etc.)
    * @param itemId the item ID to update
    * @param favorite whether the item is a favorite or not
+   * @param updatedAt optional timestamp for when the mutation occured
+   * (defaults to current server time)
    * @returns savedItem savedItem that got updated
    */
   public async updateSavedItemFavoriteProperty(
     itemId: string,
-    favorite: boolean
-  ): Promise<SavedItem> {
-    const timestamp = SavedItemDataService.formatDate(new Date());
+    favorite: boolean,
+    updatedAt?: Date
+  ): Promise<SavedItem | null> {
+    const timestamp = updatedAt ?? SavedItemDataService.formatDate(new Date());
     const timeFavorited = favorite ? timestamp : '0000-00-00 00:00:00';
     await this.db('list')
       .update({
--- a/src/models/SavedItem.ts
+++ b/src/models/SavedItem.ts
@@ -61,6 +61,54 @@ export class SavedItemModel {
   }
 
   /**
+   * 'Favorite' a Save in a Pocket User's list
+   * @param id the ID of the SavedItem to favorite
+   * @param timestamp timestamp for when the mutation occurred. Optional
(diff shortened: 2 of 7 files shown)


src/dataService/savedItemsService.ts
```diff
@@ -177,13 +177,16 @@ export class SavedItemDataService {
...
    * @returns savedItem savedItem that got updated
    */
   public async updateSavedItemFavoriteProperty(
     itemId: string,
-    favorite: boolean
-  ): Promise<SavedItem> {
-    const timestamp = SavedItemDataService.formatDate(new Date());
+    favorite: boolean,
+    updatedAt?: Date
+  ): Promise<SavedItem | null> {
```
kschelonka, line 188: again just being more explicit/correct

src/models/SavedItem.ts
```diff
@@ -60,6 +60,54 @@ export class SavedItemModel {
...
+  public async favoriteById(
+    id: string,
+    timestamp?: Date
+  ): Promise<SavedItem | null> {
+    const savedItem = await this.saveService.updateSavedItemFavoriteProperty(
+      id,
+      true,
+      timestamp
+    );
+    if (savedItem == null) {
```
Herraj, line 81: should we use `===` if we are strictly checking if it's `null` 🤔

src/models/SavedItem.ts
```diff
@@ -60,6 +60,54 @@ export class SavedItemModel {
...
+   * @returns The updated SavedItem if it exists, or null if it doesn't
+   * @throws NotFound if the SavedItem doesn't exist
+   */
+  public async unfavoriteById(id: string, timestamp?: Date) {
+    const savedItem = await this.saveService.updateSavedItemFavoriteProperty(
+      id,
+      false,
+      timestamp
+    );
+    if (savedItem == null) {
```
Herraj, line 103: Same as above. Should be fine since I think we are handling `undefined` too in this block

Herraj (approved):

src/models/SavedItem.ts
kschelonka: thanks for the comment, typically when there isn't a behavior difference between null and undefined I use `==` because it's less prone to developer error bugs

PR #804, opened 2023-06-02
chore(docs): update README

jpetto:
## Goal

update README. there are no private packages anymore, so no need for the github token.

- remove sha from localstack docker-compose image for apple chip support
- update .env.example to reflect README changes


.env.example
```diff
@@ -1,2 +1 @@
-GITHUB_TOKEN=<YOUR_PERSONAL_TOKEN>
-PARSER_ENDPOINT=
+PARSER_DOMAIN=
```
jpetto, line 1: this is the correct env variable name referenced in code.

kschelonka (approved):

PR #805, opened 2023-06-02
feat(undelete): undelete by url

kschelonka:
To support offline mobile clients. Continuing patterns from #801, #802, etc.

[IN-1474]

[IN-1474]: https://getpocket.atlassian.net/browse/IN-1474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/dataService/savedItemsService.ts
+++ b/src/dataService/savedItemsService.ts
@@ -326,9 +326,16 @@ export class SavedItemDataService {
    * which status the item is assigned when moved from the deleted state.
    * @param itemId
    */
-  public async updateSavedItemUnDelete(itemId): Promise<SavedItem> {
+  public async updateSavedItemUnDelete(
+    itemId: string,
+    updatedAt?: Date
+  ): Promise<SavedItem | null> {
+    const timestamp = updatedAt ?? SavedItemDataService.formatDate(new Date());
     const query: any = await this.getSavedItemTimeRead(itemId);
-
+    // Does not exist or was hard-deleted
+    if (query == null) {
+      return null;
+    }
     // This is a check to determine if the saved item was previously archived. Fun, right?
     const status =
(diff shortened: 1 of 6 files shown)


src/dataService/savedItemsService.ts
```diff
@@ -326,9 +326,16 @@ export class SavedItemDataService {
...
    */
-  public async updateSavedItemUnDelete(itemId): Promise<SavedItem> {
+  public async updateSavedItemUnDelete(
+    itemId: string,
+    updatedAt?: Date
+  ): Promise<SavedItem | null> {
+    const timestamp = updatedAt ?? SavedItemDataService.formatDate(new Date());
     const query: any = await this.getSavedItemTimeRead(itemId);
-
+    // Does not exist or was hard-deleted
```
kschelonka, line 335: This would previously throw an unintuitive server error (would fail on line 341, attempting to access key on undefined). This doesn't change the end state behavior but results in a better response. The null gets caught by the calling function and is converted to NOT FOUND in business logic layer.

src/models/SavedItem.ts
```diff
@@ -94,7 +94,10 @@ export class SavedItemModel {
    * @returns The updated SavedItem if it exists, or null if it doesn't
    * @throws NotFound if the SavedItem doesn't exist
    */
-  public async unfavoriteById(id: string, timestamp?: Date) {
+  public async unfavoriteById(
+    id: string,
+    timestamp?: Date
+  ): Promise<SavedItem | null> {
```
kschelonka, line 100: Fix a missed return type

cmharlow (approved):

PR #808, opened 2023-06-05
chore(savedItem.title): deprecate savedItem.title in schema docs

sri-kirsh:
## goal

- mark `savedItem.title` as alpha as its not resolved yet
- mark them as deprecated as we will use it only for third party clients API (v3 proxy work). internal clients are expected to use `item.title`

### DIFF
--- a/schema.graphql
+++ b/schema.graphql
@@ -149,9 +149,10 @@ type SavedItem implements RemoteEntity @key(fields: "id") @key(fields: "url") {
   """
   url: String!
   """
-  The title for user saved item. Set by the user and if not, set by the parser
+  The title for user saved item. Set by the user and if not, set by the parser.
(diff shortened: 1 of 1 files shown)


schema.graphql
```diff
@@ -149,9 +149,10 @@ type SavedItem implements RemoteEntity @key(fields: "id") @key(fields: "url") {
   """
   url: String!
   """
-  The title for user saved item. Set by the user and if not, set by the parser
+  The title for user saved item. Set by the user and if not, set by the parser.
+  Supported for third party clients use-case
```
sri-kirsh, line 153:
```suggestion
  For third party clients use-case only.
```

katerinachinnappan (approved):

PR #821, opened 2023-11-01
Mysql8 compatibility

kschelonka:
## Goal
Small changes for using the mysql8-compatible library in knex ('mysql2'). The main issue is the date serialization for the special ZERO DATE timestamps we have (get converted to INVALID DATE) There's actually only one time we pull it and evaluate without converting to unix time first, and that's to check the archival status for undeleted items. Just unix-time it.


package.json
```diff
@@ -57,6 +57,7 @@
     "lodash": "4.17.21",
     "luxon": "2.5.2",
     "mysql": "2.18.1",
```
bassrock, line 59: @kschelonka Should we remove this dependency?

kschelonka: ah thank you, I thought I uninstalled already. nice catch

kschelonka: I also removed the otel library for it

bassrock (dismissed):

bassrock (approved):

nina-py (approved):

PR #825, opened 2023-11-27
chore(nov2023): updating to latest standards

bassrock:
## Goal

- Follow items in https://mozilla-hub.atlassian.net/wiki/spaces/PE/pages/470582325/2024+-+Backend+Repo+Up+To+Date+Checklist
- Update node to node 20.10.0 to resolve memory leak issue https://mozilla-hub.atlassian.net/browse/POCKET-9193
- Add in sentry request isolation to make better work of Sentry errors
- Upgrade all packages and CI

## References

Documentation:
* Project doc

PR #826, opened 2023-11-30
fix(mem): removing kinesis batcher as possible memory leaker

bassrock:
## Goal

The mem leak from https://github.com/Pocket/list-api/pull/825 continues.

I believe that it is likely the batch processor optimization that we had for Kinesis because it relied on timers and was polling. And because when removed, the tests no longer needed a forceExit param cause something was left open.

We will try deploying without the batch optimization since we should be well below kinesis limits for a good while.

Screenshot 2023-11-29 at 5 32 04 PM


bassrock:
No idea why the linter wanted these changes all of a sudden..

src/test/graphql/mutations/savedItemsMutationService-delete.integration.ts
```diff
@@ -123,6 +123,7 @@ describe('Delete/Undelete SavedItem: ', () => {
 
   afterAll(async () => {
     await writeClient().destroy();
+    await readClient().destroy();
```
bassrock, line 126: This test needed the readClient destroyed to not hang them. I think its cause we made 2 queuries, 1 thats a mutation, 1 thats not. And we have logic that switches the client based on mutations. So when the query runs, it inits the readClient in the execution context, but we never cleaned it up.

kschelonka (dismissed):

